PR #14, opened 2020-09-02
add capacity targets for electrolysis in EU and DEU for NDC

fschreyer:
Add NDC capacity targets for electrolysis in 2025 and 2030 following recent EU Commission Hydrogen Strategy and German Hydrogen Strategy (see links in code). Relevant for REMIND-EU. The EU-wide target of 40GW in the EU is disaggregated to subregions (which do not have national target) by GDP. Note: The unit is GW(input) of electrolysis, so this still needs to be multiplied by the electrolysis efficiency in the datainput code.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '3552384'
+ValidationKey: '3701400'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


christophbertram (approved):
Looks good. A useful explicit clarification could comment on the unit of the hard-coded GW numbers. I guess this is the electricity input, but can't harm to make this explicit.

PR #17, opened 2020-09-09
Add ARIADNE Population and GDP data to historical.mif

fbenke-pik:


LaviniaBaumstark (commented):
Do data from ARIADNE_ReferenceScenario already use the same names as used in our model reporting?

fbenke-pik:
> Do data from ARIADNE_ReferenceScenario already use the same names as used in our model reporting?
The values in model, scenario, variable and unit were chosen in coordination with @Renato-Rodrigues  and considering the expected format in mrremind reporting

LaviniaBaumstark (approved):

PR #21, opened 2020-09-24
Add TtW efficiency values to calculate real UE values.

Loisel:
For transport demand trajectories, at the moment REMIND uses "fossil-fuel equivalents" which is not in line with the common understanding of useful energy. Instead, I suggest to use motive energy and consequently tank-to-wheel efficiencies as conversion factors from FE to UE.

Sources for the TtW efficiencies are given directly in the source code.

Note: This requires also a change to the `generisdata_tech.prn` in REMIND, so once this is merged and there is new inputdata, the corresponding patch to REMIND should also be merged. I will link the PR as soon as it is available.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '4260060'
+ValidationKey: '4446960'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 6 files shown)


MariannaR:
It might be that also the `remind` library needs updates in this direction?

Loisel:
Which parts do you have in mind? I guess the energy service conversion factors have to be adopted. Anything else?

LaviniaBaumstark (approved):

PR #33, opened 2020-10-28
Complex transport update

MariannaR:
This PR contains:

- the substitution of the bunkers share calculation on total liquids demand: now it is based on EDGE-T values. The previous function to calculate bunkers share have been removed.

- some minor code cleaning.


christophbertram (approved):
The resulting shares are lower in the historic period (2005,10,15,20) than with the previous approach, and start decreasing from 2020 onward. This should be checked and potentially revised in future work on bunker demands

PR #37, opened 2020-10-29
Small bugfix in bunkers share calculation.

MariannaR:
Bunker shares are calculated as a share of total liquids FE demand for HDVs (not as a share of total FE demand for HDVs).


christophbertram (commented):
All good

PR #47, opened 2021-01-27
add readREMIND_EU() and update calcFEdemand() for tuned industry FE calibration data

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- REMIND-EU uses industry/fixed_shares FE calibration data tuned to
  better match European country development
- calcFEdemand() is modified to substitute previously used data
- this will change Baseline scenarios for all non-REMIND-EU runs that
  use industry/fixed_shares, too (e.g., SDP)

### DIFF
--- a/R/calcFEdemand.R
+++ b/R/calcFEdemand.R
@@ -576,7 +576,22 @@ calcFEdemand <- function(subtype = "FE") {
   getNames(reminditems) <- gsub("SDP","gdp_SDP",getNames(reminditems))
 
   if ('FE' == subtype) {
-
+    
+    # ---- _replace industry FE demand for Europe by REMIND-EU-tuned data ----
+    # read REMIND-EU data
+    replacement <- readSource(type = 'REMIND_EU',
+                              subtype = 'fixed_shares_FE_calibration_data',
+                              convert = FALSE)
+    
+    # harmonise dimension names
+    dimnames(replacement) <- setNames(
+      dimnames(replacement),
+      names(dimnames(reminditems)))
+    
+    # overwrite data
+    reminditems[getRegions(replacement),getYears(replacement),
+                getNames(replacement)] <- replacement
+    
     # ---- _modify SSP1/SSP2 data of CHN/IND further ----
     # To achieve projections more in line with local experts, apply tuning 
     # factor f to liquids and gas consumption in industry in CHN and IND. 
--- a/R/readREMIND_EU.R
+++ b/R/readREMIND_EU.R
@@ -26,17 +26,7 @@ readREMIND_EU <- function(subtype) {
                col_names = c('t', 'regi', 'SSP', 'pf', 'value'),
                col_types = 'icccd',
                comment = '*') %>% 
-        # # add useless NA data to please the all-mighty madrat god
-        # complete(nesting(t, SSP, pf), 
-        #          regi = read_delim(
-        #            file = system.file('extdata', 'iso_country.csv', 
-        #                               package = 'madrat'),
-        #            delim = ';',
-        #            col_names = 'iso3c',
-        #            col_types = '-c',
-        #            skip = 1) %>% 
-        #            pull('iso3c')) %>% 
-        as.magpie(spatial = 4, temporal = 1) %>% 
+        as.magpie(spatial = 2, temporal = 1) %>% 
         return()
     },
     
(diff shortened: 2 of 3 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@Renato-Rodrigues, @LaviniaBaumstark I left this commit unfinished (the `buildLibrary()` stuff works, but I dropped it) in case you want to coordinate somehow and pull all the changes of the merge into `mrremind` all at once.

@bs538 not sure if you anticipate to do new input data for the SDP paper, but if so, be advised that this would alter your baseline compared to earlier scenarios.

Renato-Rodrigues:
@0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q @LaviniaBaumstark We can explain the change better in the REMIND meeting before finishing the commit. 

This change replaces directly the FE industry trajectories for SSP1, SSP2 and SSP5 for EUR and NEU countries. As consequence, the SDP is also recalculated for Europe and NEU.
 
Unfortunately, I have no idea on how the SSP3 and SSP4 FE trajectories are derived, so I have no clue if we should keep the values for them as it is, or if we should also include some change to make these remaining SSPs compatible with the new EU trajectories assumptions.

In the medium-term we should have a much more consistent FE trajectory determination as soon as Michaja does his magic with the industry subsectors realization. By then we will be able to get rid of these exogenously defined FE trajectories based on REMIND-EU assumptions.

Renato-Rodrigues (approved):

bs538:
thanks for keeping me in the loop here, good to know! as long as the SDP paper is in review, there is a chance that the reviewers will make us change some part of the FE demands. can we ensure that this can be done easily despite your updates?

also, I'd like to understand the motivation better. is this fixing a problem? or just bringing over some updates from REMIND-EU? also, how big is the change to the FE demands?

Renato-Rodrigues:
@bs538 I hope my reply in the meeting was clear enough on the updates question, if not please let me know so I can try to explain it better.

On the motivation, mrremind FE industry split into European regions/countries presented "not that good" resulting FE trajectories for each countries, and were also incompatible with 2030 and mid-century levels when compared to other scenarios around.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> thanks for keeping me in the loop here, good to know! as long as the SDP paper is in review, there is a chance that the reviewers will make us change some part of the FE demands. can we ensure that this can be done easily despite your updates?

To elaborate on what Renato said during the REMIND meeting: your best shot in any case would be to modify your input data locally, since re-generating input data would get you all kinds of other changes that would alter your REMIND results.
Checking out the old version of `mrremind` (you can find the version number used in the header of your input files) _might_ give you consistent results, but if the input data changed (which still is not under version control @LaviniaBaumstark), you are s**t out of luck.
So if you need to do something like that, load your input data into R with `read_csv()` and work from there.  

But when you do the follow-on SDP paper, there will be lots of difference, some of which will be a result of this commit.
 
> also, I'd like to understand the motivation better. is this fixing a problem? or just bringing over some updates from REMIND-EU?

REMIND-EU used FE data that was tune using region/country specific knowledge for the EU countries from the ARIADNE project, so should be of higher "quality" than the original REMIND 11 assumptions that where grandfathered in.

> also, how big is the change to the FE demands?

See [REMIND-EU_merger_industry_FE_comp.pdf](https://github.com/pik-piam/mrremind/files/5887657/REMIND-EU_merger_industry_FE_comp.pdf)

bs538:
thanks both for the clarifications! follow-up to Michaja: thought this was just about EU, why do all other regions change as well? (sorry if I'm a bit slow here...)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> thought this was just about EU, why do all other regions change as well? 

image

Relevant for you are pages 8–14, the comparison between what is currently coming out out `mrremind` and what is coming out of the commit.
Pages 1–7 are for Renato, because REMIND-EU did not pull all the changes to the FE demand that were made during the SDP process.

bs538:
got it, thanks!

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@Renato-Rodrigues do you still need this?  Or should I toss it out …

PR #48, opened 2021-02-03
Change in weights for energy in transport.

MariannaR:
This PR contains the change in weights (ES demand in 2010) for energy related entries in transport.


Loisel (approved):
looks good!

PR #52, opened 2021-03-11
Jrc

fbenke-pik:
Enhance read JRC logic to include additional workbooks and sheets needed for mapping to REMIND variables


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@fbenke-pik I gave the script a scrubbing, factoring out the code, generalising it and separating the data.  I don't think it makes much sense to put the mappings in a separate file.  For one think they would have to be in some .csv format, which would mean a lot of repetition, for another the mappings could only be used with this function and the function only works with the mappings, so they are a logical unit and should therefore be a code-unit as well.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #85, opened 2021-05-25
replace toolMappingFile with toolGetMapping

pfuehrlich-pik:


tscheypidi:
I suggest that Anastasis reviews this PR as I am not familiar with the contents of mrremind

giannou:
I looked at a few cases and they look alright. Trying/looking at all of them seems a bit of an overkill. The whole pre-processing for REMIND was tested, right?

pfuehrlich-pik:
Yes, on the cluster the pre-processing still works.

giannou (approved):

PR #97, opened 2021-07-20
Bugfix for calculating NDC hydro capacities in OAS (VNM).

merfort:
Updated IRENA data lead to a bug causing NDC hydro capacity targets in VNM to be much too high. This is fixed now.
Furthermore there was some code cleaning.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '12458840'
+ValidationKey: '12482964'
 AcceptedWarnings:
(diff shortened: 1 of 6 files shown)


amnmalik:
> 
> 
> Updated IRENA data lead to a bug causing NDC hydro capacity targets in VNM to be much too high. This is fixed now.
> Furthermore there was some code cleaning.

Actually, the bug was within the code so updating IRENA data wasn't the problem. Anyway, looks solved now

PR #109, opened 2021-09-15
Create input data from EDGE-B results with varying RCP scenario.

rkrekeler:
With this PR, mrremind reads EDGE-B results that now have an RCP dimension in addition to the SSP scenario. For all existent inputfiles, only fixed climate is used giving identical results as before. `fullREMIND.R` now creates a new input file `pm_fe_demand_build.cs4r` that contains FE demands in buildings with the additional RCP dimension. This can later be used in REMIND to overwrite demands in buildings to consider specific climate change scenarios. Moreover, this PR introduces source data versioning for EDGE-B results.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '13781670'
+ValidationKey: '13974900'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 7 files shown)


LaviniaBaumstark (commented):
there is no specific handling for the version of the source needed?

LaviniaBaumstark (approved):

PR #110, opened 2021-09-16
Adapt mrremind to SSPs-EDGET scenarios match.

MariannaR:
This should make the SSPs match with the EDGE-T scenarios. To be merged only when the `edgeTransport` and `edgeTrpLib` changes are accepted.

### DIFF
--- a/R/calcEDGETrData.R
+++ b/R/calcEDGETrData.R
@@ -20,14 +20,16 @@ calcEDGETrData <- function() {
                                           y = c("SDP", "SSP1")),
                                 cartesian(x = c("ElecEra", "ConvCase", "HydrHype", "Mix"),
                                           y = c("SSP2")),
+                                cartesian(x = c("ElecEra"),
+                                          y = c("SSP2Ariadne")),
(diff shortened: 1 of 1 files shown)


Loisel (commented):
Looks good, I think it makes sense to keep all scenarios for SSP2. So it is 8 in total?

MariannaR:
Probably one extra scenario has to be included, `SSP2-lowEn`. If extra scenarios are requested for other projects they should be (relatively) easy to add :)

MariannaR:
@Loisel : maybe the REMIND calibration (as it is now designed) uses ConvCaseWise for all Wise scenarios. It is worth looking into that to see if the script here https://github.com/remindmodel/remind/blob/c5ae6d0d82e1093e065ed215ea189c5dc202b14d/scripts/start/prepare_and_run.R#L292 accepts ElecEraWise as starting value for the calibration, and change accordingly here https://github.com/remindmodel/remind/blob/c5ae6d0d82e1093e065ed215ea189c5dc202b14d/config/default.cfg#L407

PR #145, opened 2021-11-05
Add pre-commit config to automatically detect common problems

mikapfl:
This will detect (and sometimes fix) common problems in pull requests using github actions and https://pre-commit.ci .

Currently detected problems:
* file name conflicts on case-insensitive file systems
* invalid JSON
* invalid YAML
* unresolved merge conflicts
* byte-order markers

Additionally, we could check for missing EOF or whitespace at the end of lines; however, there are a lot of these at the moment in mrremind, so you would have to discuss if you want to clean that up or just leave it as it is for now. The downside of cleaning it up would be that it creates some merge conflicts. Depending how much development is happening in parallel at the moment, that is potentially quite some work. On the other hand, added whitespace and missing EOFs do sometimes lead to subtle errors which can also be quite a lot of work detecting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.0.1
+    hooks:
+    -   id: check-case-conflict
+    -   id: check-json
+    -   id: check-yaml
+    -   id: check-merge-conflict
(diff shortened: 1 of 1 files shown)


mikapfl:
This is the same thing we already have in remind, where it is working without problems since 3 weeks.

LaviniaBaumstark (approved):

PR #162, opened 2021-12-01
fix FE aggregation bug v2

pweigmann:
Another FE aggregation bug resulting in roughly doubled FE values


fbenke-pik (approved):

pweigmann:
This bug affected the variables "FE" and "FE|Gases/Solids/Liquids"

PR #170, opened 2021-12-07
add Heat Roadmap Europe as FE source

pweigmann:
Includes new files _readHRE.R_, _convertHRE.R_ and _calcHRE.R_. 

Data is available only for some European countries, for now the rest is filled with NA, resulting in little to no data after regional aggregation.


rkrekeler (approved):
Looks good. Thank you so much for the help, @pweigmann!

PR #177, opened 2021-12-17
Input data now contains the new Mix1-4 variants from EDGE-T & version…

Loisel:
… bump.

The data for the tech focus scenarios is *not* generated anymore. Please use older input data revisions if you want to use `ElecEra`, `HydrHype` or `ConvCase`.

### DIFF
--- a/R/calcEDGETrData.R
+++ b/R/calcEDGETrData.R
@@ -15,7 +15,7 @@ calcEDGETrData <- function() {
   ## for all "default" SSP variants we ship the whole zoo of EDGE-T scenarios
   edgetScenarios <- strsplit(cartesian(
     c("SSP1", "SSP2", "SSP5", "SSP2EU", "SDP"),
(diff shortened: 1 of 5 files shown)


R/calcEDGETrData.R
```diff
@@ -15,7 +15,7 @@ calcEDGETrData <- function() {
   ## for all "default" SSP variants we ship the whole zoo of EDGE-T scenarios
   edgetScenarios <- strsplit(cartesian(
     c("SSP1", "SSP2", "SSP5", "SSP2EU", "SDP"),
-    c("ElecEra", "HydrHype", "Mix", "ConvCase")), split=".", fixed=TRUE)
```
LaviniaBaumstark, line 18: why mix1-4, are there any meaningful names possible? maybe for later...

Loisel: those are intended as a temporary solution, so I used these dummy names (I agree with you)

LaviniaBaumstark (approved):

Issue #194, opened 2022-01-20
Nonsensical regional detail in historical.mif


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
This
https://github.com/pik-piam/mrremind/blob/222bffaf400372b7e2a3d05f10ba0cab1e598fcb/R/fullREMIND.R#L167
writes `Production|Industry|Cement` and `Production|Industry|Steel` (and possibly others) in regional detail to `historical.mif`, when the source ([`readIEA(subtype = 'industry')`](https://github.com/pik-piam/mrremind/blob/master/R/readIEA_ETP.R)) only contains global data.  The disaggregation using GDP figures and re-aggregation lead to nonsensical data that is distracting and misleading during validation.
![image](https://user-images.githubusercontent.com/53254462/150376223-c5492a25-3e90-470a-b28c-c118da385888.png)
Is there a way to have _only_ global data for these variables in the file?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Tagging @fbenke-pik since he seems to be involved with redesigning `historical.mif`.

fbenke-pik:
Thanks for bringing this up, @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q 

This issue comes up in other sources as well and I had to exclude valid data from historical.mif due to this problem before, e.g. IEA WEO 2021 free data set only has global data for most of the variables.

In consequence, we cannot include data on GLO level, because the disaggregation on country-level produces inaccurate data we do not want in historical.mif. So would be great to be allowed to include data only on global level for **some** variables in calcOutput.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I think a mix of global and regional values in a single `calcOutput` does not work, but you can have a calc-function for global data and one for regional data and for the data you want to use only on global level you can then run `calcOutput(..., aggregate="glo")`

So `calcHistorical()` would call `calcOutput('Historical_global', …, aggregate = 'glo')` and `calcOutput('Historical_regional', …)`?  And by `calcOutput('Historical', …)` we would get on `magclass` object with lots of `NA`s for the global data in the regions, which would also get written to `historical.mif`?

fbenke-pik:
This has been solved via custom aggregation functions whenever needed, a standard functionality provided by madrat
For example: https://github.com/pik-piam/mrremind/blob/master/R/calcIEA_ETP.R#L95

PR #205, opened 2022-02-03
Generate global and regional mif

fbenke-pik:
Following a discussion with @LaviniaBaumstark , we are suggesting the following solution to issue #194 

From now on, we create 3 historical.mif files: in addition to the one with any regional resolution, we also have on with only regional resolution and one with only global resolution. These have to be stitched together again in remind2 compareScenarios. 

Use cases demonstrated in this PR:
1. IEA ETP data for Industy Production of Steel and Cement is accurate on global level, but not on regional level due to lacking regional granularity of the source. With the proposed solution, we can now remove the regional data in mrremind without losing the global data. 
2. The majority of IEA WEO 2021 data is only available for region GLO and regional disaggregation is too inaccurate. Therefore only the GLO data is used and goes into global historical.mif. A small subset of the data is also available for some of the regions, this goes into the regional historical.mif.

@LaviniaBaumstark Is this implementation what you had in mind?

 @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q Do you think this is a feasible solution for your problem with IEA ETP data? Advantage over filtering in remind2 is that we define in one central place which regional data we do not want to export instead of doing it in all the places where the mif files are read in. Disadvantage is that we need to read in three files now.


LaviniaBaumstark (changes requested):
I would leave the calcHistorical for all data but the data for which we do not have meaningful regional data. For those I would use your new calcHistoricalGlobal. But I would not calculate an additional file containing the regional data. They are anyway meaningless. Otherwise we could keep/add them in the default calcHistorical

fbenke-pik:
@LaviniaBaumstark If I understand you correctly, you don't see the need for calcHistoricalRegion. Right now, I am not sure how to avoid using three files. 

Using the example of IEA_ETP data, if I moved it back into calcHistorical, we would get duplicate data in historical.mif and historical_global.mif, e.g the IEA ETP data for region GLO would be in both files When merging the mifs to run compareScenarios, you would have to take care of such duplicates. If I have to write logic to filter out data from the combined mifs later on, we are back to where we are right now.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Do you think this is a feasible solution for your problem with IEA ETP data? Advantage over filtering in remind2 is that we define in one central place which regional data we do not want to export instead of doing it in all the places where the mif files are read in. Disadvantage is that we need to read in three files now.

It would "solve" the problem alright, but that is solved already.  
Looking forward I would argue this is just adding more convolutions to a 
labyrinthian system.

The root cause is `madrat`s pointless insistence on full regional detail, when 
there is none.  I'd rather that would be addressed.

Regarding this merge request, I'm more opposed than in favour.  It does 
something I currently don't need, but it will make everything more complicated
over time.

fbenke-pik:
After talking again to Lavinia, we found a solution that does not require more than one historical.mif while filtering the regional data we do not want to include using the append attribute of calcOutput. This implementation is more in line with the [magclass counterpart ](https://github.com/pik-piam/mrvalidation/blob/master/R/fullVALIDATION.R ).

LaviniaBaumstark (approved):

LaviniaBaumstark (approved):

PR #211, opened 2022-02-15
"EDGE-Industry"

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Draft pull request so you don't get a heart attack when it hits.

I put all EDGE-Industry stuff into one file.  It brings the calculations of industry demand trajectories (cement, chemicals, steel, otherInd) into the package and properly integrates them with upstream projections (e.g. GDP) which did change in the past.
The functions have a lot of stuff in them (returning interim results, saving plots) that is not needed for `mrremind`, but for my paper.  Otherwise I would have to maintain two code bases, which is simply too much work.
I tested 277d662 (`/p/tmp/pehl/preprocessing-remind/log-22841280.out`) and will look the results (`f_fedemand.cs4r`) over.  Further changes will trickle in over the coming days.  I suggest that whatever is in gets into the next calibration/validation round, whatever is not gets into the next, and so on.


R/EDGE-Industry.R
```diff
@@ -0,0 +1,2518 @@
...
+#' @importFrom quitte calc_mode duplicate duplicate_ list_to_data_frame 
+#'   madrat_mule order.levels sum_total_
+#' @importFrom readr write_rds
+#' @importFrom stats nls SSlogis sd
+#' @importFrom tidyr pivot_longer pivot_wider
+#' @importFrom zoo na.approx rollmean
+
+#' @rdname EDGE-Industry
+#' @export
+calcSteel_Projections <- function(subtype = 'production',
```
LaviniaBaumstark, line 48: why don't you call this file calcSteel_Projections? This would ensure full functionality from madrat

R/calcpm_delta_kap_industry.R
```diff
@@ -0,0 +1,31 @@
...
+#'   
+#' @author Michaja Pehl
+#'
+#' @seealso [`calcOutput()`]
+#' 
+#' @importFrom magclass new.magpie
+#' 
+
+#' @export
+calcpm_delta_kap_industry <- function() {
```
LaviniaBaumstark, line 15: is this a calc* function? or a tool* function?

R/EDGE-Industry.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: What has `madrat` functionality to do with the file name?

R/calcpm_delta_kap_industry.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: It is a "calc" function.

R/EDGE-Industry.R
LaviniaBaumstark: e.g. check of functions for caching

R/calcpm_delta_kap_industry.R
LaviniaBaumstark: ah, I see....

R/EDGE-Industry.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: And how is that, pray tell?  How does the name of a file, from which a package is compiled, influence either content or functionality of said package?

LaviniaBaumstark: based on the name it is checked whether the information from the cache can be loaded or it needs to be updated (if not started with an empty cache

LaviniaBaumstark:
the renaming could be done later

LaviniaBaumstark (approved):

PR #272, opened 2022-05-30
shift CHP values for better consistency

fbenke-pik:
Changes as requested by @robertpietzcker 

1. for each product: check if the flow "HEMAINC" is >0, if yes, do nothing; if no, add the value of the flow "ELMAINC" to "ELMAINE" and afterwards set ELMAINC to zero.
2. for each product: check if the flow "HEAUTOC" is >0, if yes, do nothing; if no, add the value of the flow "ELAUTOC" to "ELAUTOE" and afterwards set ELAUTOC to zero.

this shifts all values that are reported as CHP to pure electric plants whenever there is no heat output recorded from these plants. 

The hope would be that this solves the issues by removing CHP plants from regions that don't report heat output for these plants


merfort:
Thanks a lot! Looks good.
Just for reference, the resulting changes in secondary energy output of technologies that produce electricity and/or heat:
bioigcc
biochp
pc
coalchp
x_gas2elec
gaschp
dot

In particular there is now more `seel` attributed to `bioigcc`, `pc` and `x_gas2elec` instead of `biochp`, `coalchp` and `gaschp`, respectively, for some regions. This is only a shift in attribution, so no `seel` output gets lost. 
As a side effect there is now slightly more `seel` attributed to `dot`, since there is no `oilchp` in REMIND. This fraction of `seel` thus got lost before, now it is attributed to `dot` instead.

PR #273, opened 2022-06-01
New EDGE-B trajectories (1.07) and bugfix in alternative RCP buildings demands

robinhasse:
New EDGE-B trajectories: 1.06 -> 1.07
- better insulation projection in DEU in `SSP2 `and `SSP2EU`
- lower/higher long-term share of gas/electricity in space heating in DEU in `SSP2 `and `SSP2EU`

Bugfix in alternative RCP demands
- The files with FE and UE demands for alternative RCP scenarios in buildings had a wrong identifier for no climate change (has to be `none`. This is now fixed.
- The regex to check the result of `calcFEdemand` is adjusted to cope with "new" carriers in buildings

Other: updated my last name


robinhasse:
@LaviniaBaumstark, I checked again and this PR does not require an update of REMIND. Instead, REMIND should no longer fail for runs with alternative RCP building demands (which only I am using anyway).

LaviniaBaumstark (approved):

PR #282, opened 2022-06-14
Read number of households and buildings from IDEES

robinhasse:
`readJRC_IDEES` now reads the sheet on number of households/number of buildings from the Residential/Tertiary workbooks.

Additionally, I would like to move the entire function to `mrcommons`, as I will make use of it in `mredgebuildings`. @fbenke-pik  and @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, is anything wrong with that or anything I have to consider?

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '25492643'
+ValidationKey: '25861950'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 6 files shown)


fbenke-pik:
Looks good to me. I assume, you tested this locally?

If you want to move this to mrcommons, you definitely must move read and convert together, but you probably can leave the calc function in mrremind. I don't see any problems with that, but I would also double check and ask @LaviniaBaumstark.

robinhasse:
> Looks good to me. I assume, you tested this locally?

Sure.

> If you want to move this to mrcommons, you definitely must move read and convert together, but you probably can leave the calc function in mrremind. I don't see any problems with that, but I would also double check and ask @LaviniaBaumstark.

Of course, calc and read would need to be in the same package. I will ask Lavinia.

LaviniaBaumstark:
for me it is fine moving this to mrcommons. That's what this package is for.

fbenke-pik (approved):

PR #288, opened 2022-06-24
Rename EDGAR6 and CEDS emission variables containing NOx using NOX in…

fbenke-pik:
…stead

This renames all EDGAR6 and CEDS Emissions Variables of the pattern "Emi|NOx|.." to "Emi|NOX" to match the naming in REMIND.


chroetz (commented):
It looks correct. But I know next to nothing about mrremind, so I cannot really judge

fschreyer (approved):

PR #291, opened 2022-06-29
fix bug in IEA data aggregation causing too low steel energy share

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Aligns aggregation of IEA data with other users (coke oven and blast furnace losses all go into the steel sector) and fixes a bug contained therein.  Increases steel FE demand, lowers other subsectors FE demand, see attached [PDF](https://github.com/pik-piam/mrremind/files/9009205/REMIND_rev_comp.pdf) for comparisons.  Global peak energy demand in `SSP2EU` does not change significantly so no changes of &alpha; factors is needed.
Additional fixes regarding IEA data holes are still coming, that should deal with low specific energy demand across sectors in various regions (e.g. `MEA` and `NEN`).

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '26487580'
+ValidationKey: '26508142'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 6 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Bump @fschreyer.

PR #303, opened 2022-08-05
Add DEM_scenario column to EDGE-T inputs

Loisel:
Adds an additional dimension to EDGE-T inputs. This requires version 0.19.0 of EDGE-T to be installed.


LaviniaBaumstark (commented):
this change requires adjustments in REMIND GAMS code to deal with the new results, right?

Loisel:
yes, as soon as new inputdata is available, REMIND has to be updated as well. In fact there is a catch: we have to update edgeT to produce new inputdata with the new version of mrremind. Once we update edgeT, REMIND will stop working until we update REMIND. But if we update REMIND right away, it will break due to the unavailable input data.

LaviniaBaumstark:
sure, please, just prepare REMIND changes so that the time of inconsistency is as short as possible. thanks

LaviniaBaumstark (approved):
nice that you introduced the function compress_magpie

LaviniaBaumstark (approved):

PR #304, opened 2022-08-10
Speed up carrier remapping in calcFEdemand.

robinhasse:
The carrier aggregation in `calcFEdemand` used to loop over a list of carriers with many duplicates. This PR removes those with significant runtime improvements.

# No impact on results
I ran `calcOutput("FEdemand")` for all subtypes with `forcecache = TRUE, ignorecache = "calcFEdemand"` and made sure that the files created are identical.

# Runtime improvement
Runtime improves for all subtypes. Some buildings-specific files have higher dimensionality and therefore more duplicates. Here, runtime savings are most significant.

grafik

### DIFF
--- a/R/calcFEdemand.R
+++ b/R/calcFEdemand.R
@@ -581,11 +581,11 @@ calcFEdemand <- function(subtype = "FE") {
 
     datatmp <- data
     # Take the names of reminditems without scenario and rcp dimension
-    names_NoScen <- lapply(getNames(reminditems), function(name) {
+    names_NoScen <- unique(lapply(getNames(reminditems), function(name) {
       strsplit(name, ".", fixed = TRUE)[[1]] %>%
         `[`(!tail(getSets(reminditems), -2) %in% c("scenario", "rcp")) %>%
         paste(collapse = ".")
-    })
+    }))
 
     for (reminditem in names_NoScen) {
       # Concatenate names from mapping columns so that they are comparable with
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '26723616'
+ValidationKey: '26765102'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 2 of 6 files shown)


R/calcFEdemand.R
```diff
@@ -581,11 +581,11 @@ calcFEdemand <- function(subtype = "FE") {
 
     datatmp <- data
     # Take the names of reminditems without scenario and rcp dimension
-    names_NoScen <- lapply(getNames(reminditems), function(name) {
+    names_NoScen <- unique(lapply(getNames(reminditems), function(name) {
       strsplit(name, ".", fixed = TRUE)[[1]] %>%
         `[`(!tail(getSets(reminditems), -2) %in% c("scenario", "rcp")) %>%
         paste(collapse = ".")
-    })
+    }))
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 588: Why not
```
    names_NoScen <- lapply(unique(getNames(reminditems)), function(name) {
      strsplit(name, ".", fixed = TRUE)[[1]] %>%
        `[`(!tail(getSets(reminditems), -2) %in% c("scenario", "rcp")) %>%
        paste(collapse = ".")
    })
```
Seems more intuitive to me …

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Runtime improves for all subtypes. Some buildings-specific files have higher dimensionality and therefore more duplicates. Here, runtime savings are most significant.

Out of curiosity: are we talking seconds or minutes here?

R/calcFEdemand.R
robinhasse: Because this way, the duplicates remain. The function within lapply removes ssp and rcp from the 3rd madrat dimension and leaves the carriers. This list of carriers has the duplicates. We only need unique carriers.

robinhasse:
> > Runtime improves for all subtypes. Some buildings-specific files have higher dimensionality and therefore more duplicates. Here, runtime savings are most significant.
> 
> Out of curiosity: are we talking seconds or minutes here?

Minutes! Might be close to an hour for the entire preprocessing. Running all subtypes took less than 7 min compared to 39 min  before. I suppose you would multiply this by the number of regionmappings you calculate to estimate the overall saving...

R/calcFEdemand.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: kk

PR #310, opened 2022-08-19
Correct mapping

johannah-pik:


robinhasse (approved):
Changes look good to me. A more general remark: `readEDGETransport` does not look like a read but a calc function to me. The function does not read any files but calls other calc functions. Therefore, the roxygen header does not fit. It is either outdated or a copy paste error.

PR #314, opened 2022-08-25
read IEA 2022 data

fbenke-pik:


LaviniaBaumstark (commented):
this requires a specific mrcommons version. could you add this?

fbenke-pik:
i added the mrcommons version dependency

PR #316, opened 2022-09-01
SSP2_lowEn for all standard scenario combinations

johannah-pik:


robinhasse (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (commented):
This would make Transport use `SSP2EU_lowEn`, while the rest of REMIND uses `SSP2_lowEn` (naturally breaking things).
I do remember that @LaviniaBaumstark wanted to move to `SSP2EU_lowEn`.  Maybe it is time to do that now?

johannah-pik:
Actually not. EDGE-T would use SSP2EU_lowEn as well. 
It is a bit confusing now because we have only two demScens in EDGE-T. "Default" with no demand adjustments after the SA demand regression and "SSP2EU_lowdem". So having own demScens for EDGE-T we need to copycreate them to get the right combinations for REMIND (which is a bit annoying and happens here https://github.com/pik-piam/mrremind/blob/176c14ba33b3c4666ab1b9309775af6218ff45f6/R/readEDGETransport.R#L61). In EDGE-T there is no difference between SSP2 and SSP2EU. We should definitely discuss, if thats the right way to go or if we can find a better solution!

PR #320, opened 2022-09-27
Avoid duplication of scenarios due to imprecise filtering.

Loisel:
Some of the scenarios would be duplicated in the old version.

The new version yields the following combinations:
Screenshot from 2022-09-28 11-55-38 354670347


johannah-pik (approved):
Schaut gut aus

PR #331, opened 2022-11-16
include ODYM_RECC industry trajectories for SDP_x

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


rauner:
Do we really want this feature, I guess for SHAPE we opted for either sticking to what we have or a time variable material eff implementation.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
image
Why?  What?  Why?
… whatever.

PR #335, opened 2022-12-19
updated cement production trajectories

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Updated cement production trajectories including historic data up to 2020.  New, "less collapsing" behaviour for `CHA`.  Ideally this is included with the IEA update, so we upset the system only once.

image

image

@nicobauer, might be of interest to you.


fbenke-pik (approved):

nicobauer:
[From the zoom chat.]
I would not exclude the possibility that the regression is problematic when it comes to extrapolation. It could simply be that cement use is supported by policies and a real estate bubble that is out of proportion. If these cement boosting causes vanish, the drop would be even steeper. This is not an argument against the regression analysis as such; this is an issue about using regressions for extrapolations, if some factors are not considered.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> With four parameters I can fit an elephant, and with five I can make him wiggle his trunk.

—John von Neuman

We can add a "is this China" parameter to the regression/projection, actually we did.  But now we have to decide what the "correct" future behaviour is.
I'm open to suggestions.

nicobauer:
Every regression is subject to variable omission bias and the biased parameters, for all extrapolations we have the problem of parameter stability. John von Neumann was surely a funny guy, but he never was faced with bubble market (and I think that the real estate market in China is currently in such a situation).

nicobauer:
Ohh... a country dummy vairable will not solve the issue.

PR #337, opened 2023-01-03
Update IEA WEO 2021 data

fbenke-pik:
This PR upgrades the IEA WEO 2021 reference data from the free to paid version.


LaviniaBaumstark (approved):

fbenke-pik:
@0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q This change will also affect [calcFEdemand](https://github.com/fbenke-pik/mrremind/blob/master/R/calcFEdemand.R#L1297). It should not crash with the new data. Do we need further checks?
[data_diff.zip](https://github.com/pik-piam/mrremind/files/10383601/data_diff.zip)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> This change will also affect [calcFEdemand](https://github.com/fbenke-pik/mrremind/blob/master/R/calcFEdemand.R#L1297). It should not crash with the new data. Do we need further checks?

Well, if we don't do checks, we will do involuntary checks …

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #344, opened 2023-01-13
add NAV_lce scenario

robinhasse:
This adds another NAVIGATE scenario for energy demand in buildings. I guess @johannah-pik would have to add this scenario for transport too. So this PR should probably not be merged before this addition.


johannah-pik (approved):
Thanks Robin! :)

robinhasse:
I close this merge request [#347](https://github.com/pik-piam/mrremind/pull/347) includes these plus further changes.

PR #353, opened 2023-01-27
add near-term China historical data of gas power plants

cchrisgong:
This PR is to replace the hard coded input data in this PR: https://github.com/remindmodel/remind/pull/1194

### DIFF
--- a/R/calcCapacity.R
+++ b/R/calcCapacity.R
@@ -96,21 +96,37 @@ calcCapacity <- function(subtype) {
     WEOcap <- rename_dimnames(WEOcap, dim = 3, query = mapping, from = "WEO_techs", to="REMIND_techs")
     WEOcap <- WEOcap * 1E-03 # converting GW to TW
     
-#*** CG: near term China gas data - creating empty lines to be filled by hand below
-    CHA.2020.GasData <- new.magpie("CHN",
-                         years = getYears(IRENAcap),
-                         names = "gaschp",
-                         fill=0)
+    #    ***CG: fix CHA gas power capacities: 97 GW by September 2020 (Oxford Institute for Energy Studies:
+    #    Natural gas in China’s power sector: Challenges and the road ahead 
+    #    (https://www.oxfordenergy.org/wpcms/wp-content/uploads/2020/12/Insight-80-Natural-gas-in-Chinas-power-sector.pdf)
+    #    >60% gas plants are co-generation, rest are peaking
+    #    *** for 2018-2022, take 90GW, 90GW*0.6=54, the rest is split between ngcc and ngt
+    
+    CHA.2020.GasData <- as.magpie(
+      tribble(
+        ~region,   ~year,   ~data,      ~value,
+        "CHN",     2010,    "gaschp",   0.022,
+        "CHN",     2015,    "gaschp",   0.05,
+        "CHN",     2020,    "gaschp",   0.054,
+        "CHN",     2010,    "ngcc",     0.001,
+        "CHN",     2015,    "ngcc",     0.005,
+        "CHN",     2020,    "ngcc",     0.01,
+        "CHN",     2010,    "ngt",      0.003,
+        "CHN",     2015,    "ngt",      0.016,
+        "CHN",     2020,    "ngt",      0.026))
     
     # merge IRENA, Openmod and WEO capacities data
-    output <- new.magpie(cells_and_regions=unique(c(getRegions(IRENAcap),getRegions(Openmodcap),getRegions(WEOcap))),
-                         years = unique(c(getYears(IRENAcap),getYears(Openmodcap),getYears(WEOcap))),
+    output <- new.magpie(cells_and_regions=unique(c(getRegions(IRENAcap),getRegions(Openmodcap), getRegions(WEOcap), getRegions(CHA.2020.GasData))),
+                         years = unique(c(getYears(IRENAcap),getYears(Openmodcap),getYears(WEOcap), getYears(CHA.2020.GasData))),
                          names = unique(c(getNames(IRENAcap),getNames(Openmodcap),getNames(WEOcap), getNames(CHA.2020.GasData))),
                          fill=0)
+    
+    output[getRegions(CHA.2020.GasData),getYears(CHA.2020.GasData), getNames(CHA.2020.GasData)] <- CHA.2020.GasData
     
     output[getRegions(IRENAcap),getYears(IRENAcap),getNames(IRENAcap)] <- IRENAcap[getRegions(IRENAcap),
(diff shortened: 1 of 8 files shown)


cchrisgong:
Hi Lavinia, I get the data now in mrremind. If looks okay I will update version number, etc. Danke~

R/calcCapacity.R
```diff
@@ -72,12 +72,19 @@ calcCapacity <- function(subtype) {
     # renaming technologies to REMIND naming convention
     WEOcap <- rename_dimnames(WEOcap, dim = 3, query = mapping, from = "WEO_techs", to="REMIND_techs")
     WEOcap <- WEOcap * 1E-03 # converting GW to TW
-
+    
+#*** near term China gas data - see below
```
robertpietzcker, line 76: maybe add a comment: "creating empty lines to be filled by hand below"?

R/calcCapacity.R
```diff
@@ -156,7 +185,23 @@ calcCapacity <- function(subtype) {
     output[getRegions(WEOcap),getYears(WEOcap),getNames(WEOcap)] <- WEOcap[getRegions(WEOcap),
                                                                            getYears(WEOcap),
                                                                            getNames(WEOcap)]
-
+    
+    #    ***CG: fix CHA gas power capacities: 97 GW by September 2020 (Oxford Institute for Energy Studies:
```
robertpietzcker, line 189: I don't really understand these scripts - just wondering why you need to input the data twice?

robertpietzcker (approved):
I assume you checked the output and it contains the right values :-)

R/calcCapacity.R
cchrisgong: > I assume you checked the output and it contains the right values :-)

yes, I did ^^

R/calcCapacity.R
cchrisgong: this was the legacy from when we created a switch for wind offshore, so there are two similar copies of input data for capacity, one with and one without wind offshore.. So both input files are created, and when offshore wind is off in REMIND, it is using a file without wind offshore in it. I thought about making the with windoffshore one default, but then I remember we probably need the switch for DIETER coupling (without the very exogenously defined offshore wind production). This is probably a tangent issue tho - What do you think?

R/calcCapacity.R
cchrisgong: > maybe add a comment: "creating empty lines to be filled by hand below"?

Thanks, this works

LaviniaBaumstark (approved):

R/calcCapacity.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > this was the legacy from when we created a switch for wind offshore, so there are two similar copies of input data for capacity, one with and one without wind offshore

Which is no reason to define identical data twice in two different locations.  Define it once, then drop it into `output` at two points.  Else you will happily change the data for one file at some point in the future, wondering why your change does not affect the other file, which you are looking at.

cchrisgong: thanks - I cleaned up the redundant scripts - could you check quickly? thx

R/calcCapacity.R
```diff
@@ -72,81 +95,19 @@ calcCapacity <- function(subtype) {
...
-    WEOcap <- rename_dimnames(WEOcap, dim = 3, query = mapping, from = "WEO_techs", to="REMIND_techs")
-    WEOcap <- WEOcap * 1E-03 # converting GW to TW
-
+    
     # merge IRENA, Openmod and WEO capacities data
     output <- new.magpie(cells_and_regions=unique(c(getRegions(IRENAcap),getRegions(Openmodcap),getRegions(WEOcap))),
                          years = unique(c(getYears(IRENAcap),getYears(Openmodcap),getYears(WEOcap))),
-                         names = unique(c(getNames(IRENAcap),getNames(Openmodcap),getNames(WEOcap))),
+                         names = unique(c(getNames(IRENAcap),getNames(Openmodcap),getNames(WEOcap), getNames(CHA.2020.GasData))),
                          fill=0)
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 109: So you define some magpie object, leave it empty, to get it's "names" (one string) from it?

I'd suggest
```
    CHA.2020.GasData <- as.magpie(
      tribble(
        ~region,   ~year,   ~data,      ~value,
        "CHA",     2010,    "gaschp",   0.022,
        "CHA",     2015,    "gaschp",   0.05,
        "CHA",     2020,    "gaschp",   0.054,
        "CHA",     2010,    "ngcc",     0.001,
        "CHA",     2015,    "ngcc",     0.005,
        "CHA",     2020,    "ngcc",     0.01,
        "CHA",     2010,    "ngt",      0.003,
        "CHA",     2015,    "ngt",      0.016,
        "CHA",     2020,    "ngt",      0.026))
…
    output[getRegions(CHA.2020.GasData),getYears(CHA.2020.GasData),
           getNames(CHA.2020.GasData)] <- CHA.2020.GasData
```

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
The linter will not be happy with your changes.  But since Lavinia doesn't care …

R/calcCapacity.R
```diff
@@ -96,43 +96,44 @@ calcCapacity <- function(subtype) {
...
     output[getRegions(IRENAcap),getYears(IRENAcap),getNames(IRENAcap)] <- IRENAcap[getRegions(IRENAcap),
                                                                                    getYears(IRENAcap),
                                                                                    getNames(IRENAcap)]
+    
     output[getRegions(Openmodcap),getYears(Openmodcap),getNames(Openmodcap)] <- Openmodcap[getRegions(Openmodcap),
                                                                                            getYears(Openmodcap),
                                                                                            getNames(Openmodcap)]
     output[getRegions(WEOcap),getYears(WEOcap),getNames(WEOcap)] <- WEOcap[getRegions(WEOcap),
                                                                            getYears(WEOcap),
                                                                            getNames(WEOcap)]
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 135: You changed the order in which data is copied to `output`.   `CHA.2020.GasData` goes first, and then `IRENAcap`, `Openmodcap`, and `WEOcap` all get a chance to overwrite it.
Unless you are sure none of them contain any gas power capacities, that might return something unexpected.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #356, opened 2023-02-24
Update NDC Inputs to 2023

sophiefuchs:
- Update 2022 NDC input data with submissions until December 31, 2022.
- New 2023 input data with submissions until February 24, 2023
- Great Britain is not in EUR target as of 2022
- Added consistency checks in readUNFCCC_NDC.R


orichters (approved):
Note: Australia has strengthened its NDC target, so the higher CO2 prices / lower emissions in CAZ is plausible.

PR #357, opened 2023-03-01
consistency check in readUNFCCC_NDC.R (cond. NDC target more stringent than uncond.)

sophiefuchs:
- Added consistency check in readUNFCCC_NDC.R
- Show function "input2[2]" (readUNFCCC_NDC.R, l.92) split over two lines to avoid linter warning (120 characters)

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '30088600'
+ValidationKey: '30115767'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "mrremind: MadRat REMIND Input Data Package",
-  "version": "0.155.0",
+  "version": "0.155.1",
   "description": "<p>The mrremind packages contains data preprocessing for the\n    REMIND model.<\/p>",
   "creators": [
     {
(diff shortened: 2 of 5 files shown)


orichters (approved):
Thanks a lot.

RahelMA (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
I can't help but wonder about this check.
If the "unconditional target" is "more stringent than" the "conditional" one, does that mean the data is inconsistent?
Are users to react to these warnings?  Are they supposed to see them?
"Normal" input data preprocessing generates hundreds of warnings
```
$ tar -Oxf /p/projects/rd3mod/inputdata/output/rev6.331_62eff8f7_remind.tgz ./diagnostics.log | grep -ci warning
309
```
if the transport data is calculated and not loaded from cache, we are speaking of tens of thousands of warnings
```
$ tar -Oxf /p/projects/rd3mod/inputdata/output/rev6.4personal-cache_62eff8f7_remind.tgz ./diagnostics.log | grep -ci warning
18072
```
So if this is relevant for users in any way, issuing a warning is not the way to go.

orichters:
Yes, @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, this warning is raised if the data is inconsistent. The only user who should ever see this is the one who did an error while manually updating the NDC target xlsx file and then generates the NDC input files manually. An update of the NDC targets as in [this PR](https://github.com/pik-piam/mrremind/pull/356) should of course only happen if no warnings are raised.

sophiefuchs:
Thanks for the heads-up about usefulness of warnings anyway, Michaja. But as Oli points out, for the person updating NDC input data this hopefully will be helpful.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
:+1:

PR #364, opened 2023-03-09
Fix calls of calcGDP and calcPopulation

johanneskoch94:
Refactor (some) of the code

Specifically trying to clean up the use of dependencies, i.e. use :: explicitly, and move rarely used packages into the Suggests field along with use of rlang::check_installed().

Move readPWT back to mrdrivers


LaviniaBaumstark (approved):

johanneskoch94:
The checks didn't succeed because the newest versions of mrcommons and mrdrivers are not yet on the rse server. Due to time pressure, I will merge anyways. (No issues encountered when I tested, and none expected now.)

PR #365, opened 2023-03-13
regionally differentiated convergence parameters for specific cement production in calcIndustry_Value_Added()

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Replaces hard-coded numbers with region-specific parameters read from `/p/projects/rd3mod/inputdata/sources/ExpertGuess/cement_production_convergence_parameters.csv`.


R/EDGE-Industry.R
```diff
@@ -216,25 +221,43 @@ calcSteel_Projections <- function(subtype = 'production',
     mutate(year = as.integer(.data$year))
 
   ## historic population ----
-  population_history <- calcOutput(type = 'PopulationPast', PopulationPast = 'UN_PopDiv', aggregate = FALSE) %>%
-    tibble::as_tibble() %>%
-    dplyr::select("iso3c", "year", "population" = "value") %>%
-    quitte::character.data.frame() %>%
-    dplyr::mutate(population = .data$population * 1e6)
+  population_history <- calcOutput(type = 'PopulationPast',
```
LaviniaBaumstark, line 224: pleas keep package::function()

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Why?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@LaviniaBaumstark bump

LaviniaBaumstark (approved):

R/EDGE-Industry.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: 1. This code worked for months.  If there is a collision, it was introduced somewhere else and should be fixed there.
2. If find it hard to believe (to put it politely) that there are any name collisions for
   - `broom::tidy()`
   - `car::logit()`
   - `dplyr::mutate()`
   - `dplyr::select()`
   - `Hmisc::wtd.quantile()`
   - `quitte::character.data.frame()`
   - `quitte::magclass_to_tibble()`
   - `quitte::seq_range()`
   - `tibble::as_tibble()`
   - `zoo::na.approx()`
   - `zoo::rollmean()`
3. Half of these occurrences are for functions that are still imported into the namespace, so the `::` syntax doesn't do anything.
4. Moving packages from `Imports` to `Suggest` and checking for them with `rlang::check_installed()` for _non-optional code_ is in violation of the [CRAN rules for writing extensions](https://cran.r-project.org/doc/manuals/R-exts.html#Package-Dependencies)
   > The ‘Suggests’ field […] lists packages that are not necessarily needed.

These functions are under constant development because they are also used produce output and figures specific to papers I am writing.  I don't see why I should track useless and harmful changes made for flimsy reasons.

LaviniaBaumstark (approved):

PR #369, opened 2023-03-16
remove camelCase linter rule from this package

fbenke-pik:
Looks like lucode2 allows for per-package overwriting of the universal lucode linter rules by overwriting the lint.R files in the repo. 

@pfuehrlich-pik FYI, we want to disable this linter rule for REMIND packages only. Let me know if there is a better way to do it. 

@LaviniaBaumstark @mikapfl Thoughts on this approach? 

Btw, another variation would be
`lucode2::lintrRules(modification = list(object_name_linter(styles = c("camelCase", "snake_case"))))`

This accepts camel and snake case, but would complain for example here: `REMIND_variable`

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -1,2 +1,2 @@
-linters: lucode2::lintrRules()
+linters: lucode2::lintrRules(modification = list(object_name_linter = NULL))
 encoding: "UTF-8"
--- a/tests/.lintr
+++ b/tests/.lintr
@@ -1,2 +1,2 @@
-linters: lucode2::lintrRules(allowUndesirable = TRUE)
+linters: lucode2::lintrRules(allowUndesirable = TRUE, modification = list(object_name_linter = NULL))
 encoding: "UTF-8"


LaviniaBaumstark:
this would mean that we need to change linter rules for all REMIND packages. This is o.k. if there is no solution, to define a set of linter rules for REMIND-packages in lucodes2

pfuehrlich-pik (approved):
This is the best way to do it I think

mikapfl:
I'd rather have a list of REMIND packages in lucode2 and then apply the alternative linter rules automatically. There are a lot of REMIND packages, and writing a PR for each of them sounds like no fun at all.

fbenke-pik:
How many REMIND packages are we talking about. I currently have not more the 5-8 on my radar?

mikapfl:
I'd say:
* https://github.com/pik-piam/remind2
* https://github.com/pik-piam/mrremind
* https://github.com/pik-piam/edgeTransport
* https://github.com/pik-piam/mredgebuildings
* https://github.com/pik-piam/remindPypsa
* https://github.com/pik-piam/mredgetransport
* https://github.com/pik-piam/mrdieter
* https://github.com/pik-piam/edgeTrpLib
* https://github.com/pik-piam/rmndt

So, yeah, might also be manageable to just add this by hand.

pfuehrlich-pik:
Doing it by hand might be more (boring) work, but in the end I think it's easier to understand what's happening, and that's the most important thing for the long term I believe.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Also,
```
$ sed -i 's/ *# *nolint: *object_name_linter.*//' R/*.R
```

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Well, that makes one sad, orphaned `# nolint` line …

PR #370, opened 2023-03-17
add optional smoothing to calcODYM_RECC()

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Looks like this
ODYM_RECC_smoothie
and affects all scenarios/regions/subsectors.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '30514520'
+ValidationKey: '30710460'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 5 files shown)


bs538 (approved):
thanks Michaja! @rauner would be better placed to judge from the content side (but is gone now). To me it certainly looks more reasonable than before. 
Before merging, could you have a quick look at overall FE|Industry, just to make sure that we don't alter the demands too much? (If there are bigger changes, could you ping me again?)

@LaviniaBaumstark this is to hopefully fix the calibration issues in SDP_RC.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Before merging, could you have a quick look at overall FE|Industry, just to make sure that we don't alter the demands too much? (If there are bigger changes, could you ping me again?)

Nothing quick about it, needs new input data. :timer_clock:

PR #375, opened 2023-03-22
fix .csv file encoding in readGCPT() that crashed input data generation

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


stephenbi (approved):
Thanks for taking care of this Michaja. Working on the other issues now...

PR #382, opened 2023-03-24
Update Fgas trajectories with EDGAR7 data

gabriel-abrahao:
As reported in [this REMIND issue](https://github.com/remindmodel/development_issues/issues/186), the F-gas emissions trajectories in REMIND were causing big dips in industry emissions already in the first timesteps.

REMIND does not currently have any endogenous representation of F-gas emissions, but reads trajectories from a very old set of IMAGE scenarios and map them by RCP.

Before we can get a better bottom-up set of trajectories, this is an intermediate approach that addresses the near-term realism and strong dips but retains the dependency on the old IMAGE scenarios after 2050 as per @robertpietzcker 's suggestion.

This PR adds functions to read and convert new [EDGAR7 F-gas emissions data](https://edgar.jrc.ec.europa.eu/dataset_ghg70#p1) for 1970-2021, using AR6 GWP100 for CO2eq. It's then incorporated into `calcFGas` in the following way:

- A 2025 EDGAR value is extrapolated from 2010 and 2020
- The 2025-2050 is interpolated between EDGAR 2025 and the 2050 value of each RCP (SSP, etc.) from IMAGE.
- 2050-2150 are kept as in the IMAGE scenarios

This happens in `calcFGas` with the new subtype "interpolate2025", which is set to be default.

Global CO2eq F-gas emissions change like this from the old (IMAGEold) to the new (IMAGEinterp) implementations until 2070. Values are also provided for emissions of some specific gas species and groups, keeping the same standard that REMIND used before.

dummy

Regional values for RCP20 and none:

regional


strefler:
Thanks Gabriel! Looks quite good already. Except for rcp26 and rcp20, it looks like we could get closer to the old trajectories if we used 2040 instead of 2050 as a convergence value. But maybe the drop in rcp26/20 would get too steep then?

gabriel-abrahao:
> Thanks Gabriel! Looks quite good already. Except for rcp26 and rcp20, it looks like we could get closer to the old trajectories if we used 2040 instead of 2050 as a convergence value. But maybe the drop in rcp26/20 would get too steep then?

Indeed it would look much smoother if we converged in 2040, but 15 years to basically zero F-gas emissions seems pretty steep. Maybe instead of linearly interpolating between the two points we do something a bit more gradual, but still converging in 2050? I can try something.

strefler (approved):

PR #386, opened 2023-04-05
add 2025 PV USA capacity for upper bound

robertpietzcker:


.zenodo.json
fbenke-pik, line 1: did you delete this by accident, or was it automatically removed while running buildLibrary()

fbenke-pik: looks like this was replaced by CITATION.cff as part of the build routine!? should be fine, i guess

fbenke-pik (approved):
Looks good to me

.zenodo.json
robertpietzcker: accidentally removed - I have no idea about this. so if you say it is fine, I am happy :-)

the only thing I changed on purpose was the calcCapacity part, and there I simply copied Chris addition to CHA and changed it for the US

PR #393, opened 2023-04-14
update historic CHA gas capacity values 2010-2020

robertpietzcker:
shift from gaschp to ngcc/ngt based on the IEA energy balances values on CHP electricity production vs non-CHP electricity production

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '31192777'
+ValidationKey: '31213840'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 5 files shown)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (commented):
Should be a "source", instead of hard-coded data …

robertpietzcker:
> 



> Should be a "source", instead of hard-coded data …

well - I am definitely not the expert on input data, but my understanding was that there is a trade-off - that adjusting very specific numbers is ok in the calcXXX scripts if it is on a small basis. Sources should be used once it is more like 50-5000 entries or so.

fbenke-pik:
I am not sure if we have a strict policy requiring it to be a source (i.e. sth strict enough to reject the PR as it is). Since this is a modification and not newly introduced code, I am accepting now to not block potentially urgent updates. 

@LaviniaBaumstark Can you comment here what the general rule is?

fbenke-pik (approved):

robertpietzcker:
I'll merge now, and if there is a clear statement that all such values should be put into a source, I can shift it afterwards.

PR #396, opened 2023-05-12
Add Historical Industry Value Added Figures to historical.mif

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Test runs pending …

PR #397, opened 2023-05-15
Dev/datarise full validationremind

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Please no longer add new functionality to `calcHistorical`, but call the calc function directly in `fullVALIDATIONREMIND`.
> 
> (`calcHistorical` is legacy code and will be split up in the long run. […])

https://github.com/pik-piam/mrremind/pull/396#discussion_r1193816999

That seems arduous and error-prone.  For example, is it correct that `IEA_ETP` and `IEA_WEO_2021` are aggregated to `region` and `global` only, even when all other variables are aggregated to `missingH12` and `EU27_map` as well for the `21_EU11` mapping?  Also, lots of repeating of default arguments.

I suggest to use a list of default arguments that should apply to all `calcOutput()` calls, and a list of `calc`-functions to call along with their specific arguments.


R/fullVALIDATIONREMIND.R
```diff
@@ -13,46 +13,57 @@
...
 
-  calcOutput(type = "IEA_ETP", aggregate = "region", file = valfile,
-             append = TRUE, na_warning = FALSE, try = TRUE, isValidation = TRUE)
+  # list `calc` functions to call ----
+  # list of all `calc` functions to call, as well as their specific arguments
+  # arguments can be empty (`list()`), and can overwrite default arguments if
+  # needed
+  historical_data_calls <- list(
+    'Historical'   = list(round = 5),
+    'IEA_ETP'      = list(isValidation = TRUE),
```
fbenke-pik, line 51: IEA_ETP and WEO_2021 are special cases, where setting `columnsForAggregation` won't produce the expected data. You can only pass `region`and `global` in these two specific cases right now. 

Background: madrat does not have a way to retrieve different variable sets depending on the regional aggregation (IEA ETP and WEO have some variables we can only use on global level). So we must work with dirty hacks. (The alternative is to create two models, e.g IEA WEO regional and IEA WEO global with different variable sets, which clutters compareScenarios).

R/fullVALIDATIONREMIND.R
```diff
@@ -13,46 +13,57 @@
...
+    'IEA_WEO_2021' = list(isValidataion = TRUE),
+    'UNIDO'        = list(subtype = 'INDSTAT2') # industry value added
+    )
 
-  calcOutput(type = "IEA_WEO_2021", aggregate = "global", file = valfile,
-             append = TRUE, na_warning = FALSE, try = TRUE, isValidation = TRUE)
+  # call all listed functions ----
+  for (i in seq_along(historical_data_calls)) {
```
fbenke-pik, line 57: I am against it. Removing redundancy makes this part much harder to understand for the 95% of people who don't have an IT-Masters and are working with this. As this is an entry point to understanding where validation data is coming from, being accessible to everyone has priority here.

R/fullVALIDATIONREMIND.R
fbenke-pik: Yes, it is not good, but so far we haven't found a better way to select specific variables depending on intended regional granularity we want to use it for. If it breaks one day, we will just have to either include or remove these two variables no matter the aggregation

fbenke-pik: > You are excluding these variables only if aggregation equals 'region', when in fact you want to exclude them whenever aggregation is anything other than 'global'.

The problem is, aggregate could also be `region+global+FOO+BAR`.  It is passed like this into the calc function from what i can see.

fbenke-pik: > If it is a general problem that our aggregation is batshit crazy less then ideal, I would argue for a blacklist/whitelist mechanism to filter the resulting .mif file.

Yes, this would probably then have to be done separately in the libraries working with historical.mif, especially remind2's compareScenarios.

PR #399, opened 2023-05-23
move filtering of IEA ETP and WEO 2021 variables to retrieve function

fbenke-pik:
Following the discussion [here](https://github.com/pik-piam/mrremind/pull/397), this is a suggestion how to filter data we don't want in historical.mif, yet cannot easily remove in the madrat functions because the framework does not support it.

madrat does not offer a way to filter variables depending on the regional aggregation (e.g. remove data on sub-global level for specific variables, but keep data on global level). Currently, we achieve this by using a hack (using the aggregate value in calc and convert functions) that has limitations  (e.g. it does not work for aggregate = "reg+glo+..") and only works by accident.

The proposed solution is calculating all variables for all regions and filter regions for certain variables only afterwards by reading in the written historical.mif file again and applying appropriate filters for IEA ETP and by filtering before appending to historical.mif for IEA WEO 2021 data. These steps happen in the fullVALIDATIONREMIND function.


R/fullVALIDATIONREMIND.R
```diff
@@ -17,42 +17,70 @@ fullVALIDATIONREMIND <- function(rev = 0) {
...
+
+  valpath <- paste0(getConfig("outputfolder"), "/", valfile)
+
+  h <- read.report(file = valpath, as.list = F)
+
+  # remove IEA ETP Steel and Cement Production on sub-global level
+  h[, , "IEA ETP", pmatch = T][, , c(
+    "Production|Industry|Cement (Mt/yr)",
+    "Production|Industry|Steel (Mt/yr)"
+  )][setdiff(getItems(h, dim = 1), "GLO"),,] <- NA
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 83: Well, that will get inextricable real soon once other variables get added …

fbenke-pik: Not sure if there is anything we can do about it.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: :point_right: fbenke-pik/mrremind#1

R/fullVALIDATIONREMIND.R
```diff
@@ -17,42 +17,58 @@ fullVALIDATIONREMIND <- function(rev = 0) {
...
+    type = "IEA_WEO_2021", subtype = "global", aggregate = columnsForAggregation,
+    file = valfile,append = TRUE, warnNA = FALSE, try = TRUE, isValidation = TRUE
+  )
 
-  calcOutput(type = "IEA_WEO_2021", aggregate = "global", file = valfile,
-             append = TRUE, warnNA = FALSE, try = TRUE, isValidation = TRUE)
-
-  calcOutput(type = "IEA_WEO_2021", aggregate = "region", file = valfile,
-             append = TRUE, warnNA = FALSE, try = TRUE, isValidation = TRUE)
+  calcOutput(
```
fbenke-pik, line 59: An unresolved problem. calcIEA_WEO_2021 is called twice with 2 different subtypes. What we want is to remove all variables for any regions other than GLO from the calculation with subtype = "global".  And we want to remove all variables for GLO from the calculation with subtype = "region".  Cannot be resolved with the current approach, which indeed creates duplicates in the historical.mif as is.

fbenke-pik: One option is to do it here before writing to historical.mif right away.

LaviniaBaumstark (approved):

R/fullVALIDATIONREMIND.R
fbenke-pik: Actually had to implement another solution, which in the case of IEA WEO 2021 filters before appending to historical.mif.

R/calcIEA_WEO_2021.R
```diff
@@ -2,7 +2,9 @@
 #'
 #' @md
 #' @return A [`magpie`][magclass::magclass] object.
-#' @param aggregate Boolean indicating whether output data aggregation should be performed or not
+#' @param subtype Either "global" or "region". On global level, the source offers
+#' more variables than on regional level, but the data should not be used on sub-
+#' global level due to its coarse disaggregation.
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 7: That is not a `subtype` argument as used throughout the madrat system.  This is a switch, that does some additional computations when it is set to `'global'`, and nothing otherwise.

R/calcIEA_WEO_2021.R
```diff
@@ -30,7 +26,7 @@ calcIEA_WEO_2021 <- function(aggregate, isValidation = FALSE) { # nolint
 
   mapping$variable <- trimws(mapping$variable)
 
-  data <- readSource("IEA_WEO_2021", subtype = aggregate)
+  data <- readSource("IEA_WEO_2021", subtype = subtype)
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 29: `readIEA_WEO_2021()` does not take a `subtype` argument.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):

R/calcIEA_WEO_2021.R
fbenke-pik: But convertIEA_WEO_2021 does.

R/calcIEA_WEO_2021.R
fbenke-pik: It also determines the subtype to be used for `read_IEA_WEO_2021` (which is the more important effect of the parameter)

R/fullVALIDATIONREMIND.R
```diff
@@ -17,42 +17,65 @@ fullVALIDATIONREMIND <- function(rev = 0) {
...
+  weo <- weo["GLO", , ]
+  write.report(weo, file = file.path(getConfig("outputfolder"), valfile), append = TRUE)
+
+  weo <- calcOutput(
+    type = "IEA_WEO_2021", subtype = "region", aggregate = columnsForAggregation,
+    warnNA = FALSE, try = TRUE, isValidation = TRUE
+  )
+
+  weo <- weo["GLO", , invert = TRUE]
+  write.report(weo, file = file.path(getConfig("outputfolder"), valfile), append = TRUE)
```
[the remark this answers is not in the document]
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: So, question is: does `calcIEA_WEO_2021(subtype = 'region')` return different _global_ values then `calcIEA_WEO_2021(subtype = 'global')`, and if so: why?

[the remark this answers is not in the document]
fbenke-pik: >  So, question is: does calcIEA_WEO_2021(subtype = 'region') return different global values then calcIEA_WEO_2021(subtype = 'global'), and if so: why?

This is all so complicated because we want to get as much as possible from the source.
There is global data available for more variables than regional data. And the regional data does not add up to global data.

So with subtype global, we take all the global data and disaggregate it with a very rough weight (because it is a formal madrat requirement), but we never want to use and data on sub-global level. 

For the subtype region, we look at the country and region level, which is available for less variables and even and the regions + countries dont add up to global. We take all we can get, but don't want to use the global value summed up from this, because it is incomplete. 

In the end you have a set with GLO from one calculation and some regional data from another calculation. Normally, in madrat, you either can disaggregate GLO, or read in countries and madrat aggregates to GLO + other regions.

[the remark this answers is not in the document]
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: If not, then all variables from `calcIEA_WEO_2021(subtype = 'region')` can be filtered with `filter_historical_mif()`.

R/calcIEA_WEO_2021.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Ah, ok.  Getting passed on to `convertIEA_WEO_2021()`.

R/calcIEA_WEO_2021.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Ah, ok.  Getting passed on to `convertIEA_WEO_2021()`.

R/fullVALIDATIONREMIND.R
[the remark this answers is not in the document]
fbenke-pik: > So, the strategy would be to modify convertIEA_WEO_2021() to disregard subtype, have it do what it is doing under subtype == 'region', but include the global figures as under subtype == 'global', and then filter for region/variable combinations we like.

This is not how convert functions work. Per convention, a convert function must return a magclass object with the 249 regions, otherwise it throws an error.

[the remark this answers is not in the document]
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: So, it would be possible to 
- take the region/country values of the variables that are available,
- subtract them from the global totals,
- disaggregate the difference crudely across the countries for which there is no data (that way the aggregated sums would match the global values),
- and finally filter the result for global values and some regional values using `filter_historical_mif()`.

Not worth it.

To answer my question
> Why? There's a shiny new function to perform filtering like that? Why???

Because magclass.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

R/fullVALIDATIONREMIND.R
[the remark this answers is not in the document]
fbenke-pik: Yes, exactly. This would be possible, but very complicated.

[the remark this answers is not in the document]
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: I don't think it's complicated.  Probably less complicated then what is done now.  But it would be a lot of work to invest.  Maybe earmark that for refactoring.

PR #401, opened 2023-06-05
Update UNFCCC to 2023 values

fbenke-pik:
This PR reads in the latest UNFCCC values. I had to remove some values, which cannot be read in, because positions vary across source files (this has always been the case, causing bugs in the current UNFCCC data). I checked that this change does not break other madrat functions and added an alternative logic to still retrieve all values used in other madrat functions.


Renato-Rodrigues (approved):
As long as none of the removed variables are not used anywhere, it looks good to me.  
As far as I know, none of them are not relevant in any way to REMIND standalone.
Thank you!!

fbenke-pik:
I am fairly confident that the removed stuff is not used anywhere, as it is very specific. Even if it was used, it would cause bugs right now. If sth breaks still, I can easily fix it.

PR #404, opened 2023-06-13
update NDC targets to 2023-05-31

orichters:
Updated NDCs for Holy See / Vatican, Kiribati and Turkey.

Has a strange effect: Turkey strengthens its 2030 NDC target from -21% to -42%, but from a incredibly high BAU scenario (roughly four times their 2020 emissions). So high the NDC target was until now [dropped by our script](https://github.com/pik-piam/mrremind/blob/master/R/convertUNFCCC_NDC.R#L345-L347) as unrealistic. Now, it is below the threshold and actually has an impact: The strengthened target basically leads to 0 carbon taxes because the NDC target is overachieved anyway. Stupid. What should we do?

Kiribati leads to massively increased coverage in `OAS:` before just `43.90 %` of 2005 emissions had a NDC target, now it jumps to `44.02 %`.

image

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '31890778'
+ValidationKey: '31916835'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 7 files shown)


Renato-Rodrigues (approved):

LaviniaBaumstark (approved):

robertpietzcker:
I think it is fine that Turkey/NEU  now doesn't have a carbon price under NPI in the next years - if the target is too weak to have any effect, then that is an outcome

PR #413, opened 2023-07-28
Average buildings FE paths to 5-yr time steps

robinhasse:
The FE demand in buildings exists at yearly resolution for historic periods. For REMIND, we have picked every 5th year until now. To get smother trajectories and comply with our understanding that each time step represents a multi-year average, we now average the yearly historic demand in buildings to 5-year time steps.

This reduces volatility most notably for `feheb`. Below, a comparison for DEU. I did not only average EDGE-B results but also the stationary component which has yearly resolution until 2010 as it also introduced significant volatility.

grafik

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '32675220'
+ValidationKey: '32694786'
 AcceptedWarnings:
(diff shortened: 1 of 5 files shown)


robertpietzcker (approved):
I don't get the R code, but the output looks good.

PR #425, opened 2023-09-14
abort generation of validation data in case of an error

fbenke-pik:
Input data generation for the historical.mif has been failing for over 2 months and we did not notice it, because the fullVALIDATIONREMIND carried on with processing.  During that time, three bugs got into mrremind and mrcommons without us noticing it.

I changed the setting to crash in that case.


LaviniaBaumstark (approved):
thanks

dklein-pik (approved):
Great! And not great that we didnt notice

PR #429, opened 2023-10-17
update IEA EV Outlook data

fbenke-pik:
This is an update of the IEA EV Outlook data from 2021 to 2023. We get  2 more years and 3 more countries. As this is only used in historical.mif, should not break anything.


R/fullVALIDATIONREMIND.R
```diff
@@ -75,6 +75,14 @@ fullVALIDATIONREMIND <- function(rev = 0) {
   weo <- weo["GLO", , invert = TRUE]
   write.report(weo, file = valfile, append = TRUE)
 
+  ## IEA EV Outook
+
+  calcOutput(
+    type = "IEA_EVOutlook", file = valfile,
+    aggregate = columnsForAggregation, append = TRUE, warnNA = FALSE,
+    try = FALSE
```
LaviniaBaumstark, line 83: nice, that you take the chance to migrate it to fullVALIDATION.R!

LaviniaBaumstark (approved):

PR #431, opened 2023-10-18
make sure that years are not lost when reading in worldsteel

fbenke-pik:
Currently input data generation fails because years are lost when reading in worldsteel. Expects column names as numbers, but they are read in as X2005, X2010 ...

```
~ Run readSource(type = "worldsteel", convert = FALSE)
~~ WARNING: NAs introduced by coercion
~~ WARNING: NAs introduced by coercion
~~ WARNING: NAs introduced by coercion
~~ WARNING: NAs introduced by coercion
~~ WARNING: NAs introduced by coercion
~~ WARNING: NAs introduced by coercion
~~ WARNING: NAs introduced by coercion
~~ ERROR: 
~~ error in evaluating the argument 'x' in selecting a method for function 'as.magpie': error in evaluating the argument 'x' in selecting a method for function 'as.data.frame': NAs are not allowed in subscripted assignments
Error in `h()`:
! error in evaluating the argument 'x' in selecting a method for function 'as.magpie': error in evaluating the argument 'x' in selecting a method for function 'as.data.frame': NAs are not allowed in subscripted assignments
```

### DIFF
--- a/R/calcIEA_EVOutlook.R
+++ b/R/calcIEA_EVOutlook.R
@@ -51,6 +51,8 @@ calcIEA_EVOutlook <- function() {
   # set 0s in other CHA countries than China to approximate CHA as China
   x[c("HKG", "MAC", "TWN"),,] <- 0
 
+  x <- add_dimension(x, dim = 3.1, add = "scenario", nm = "historical")
+
   return(list(
     x = x,
     weight = NULL,
--- a/R/readworldsteel.R
+++ b/R/readworldsteel.R
@@ -56,7 +56,8 @@ readworldsteel <- function(subtype = 'detailed') {
           read_ods(path = file_path, sheet = sheet, na = '...') %>%
             as_tibble() %>%
             mutate(name = sheet) %>%
-            pivot_longer(c(-'country', -'name'), names_to = 'year', names_transform = list(year = as.integer))
+            pivot_longer(c(-'country', -'name'), names_to = 'year') %>%
+            mutate(year := as.integer(gsub("^X", "", year)))
         }) %>%
         bind_rows() %>%
         add_countrycode_(origin = c(country = 'country.name'),
--- a/R/readworldsteel.R
+++ b/R/readworldsteel.R
@@ -57,7 +57,7 @@ readworldsteel <- function(subtype = 'detailed') {
             as_tibble() %>%
             mutate(name = sheet) %>%
             pivot_longer(c(-'country', -'name'), names_to = 'year') %>%
-            mutate(year := as.integer(gsub("^X", "", year)))
+            mutate(!!sym("year") := as.integer(gsub("^X", "", !!sym("year"))))
         }) %>%
         bind_rows() %>%
         add_countrycode_(origin = c(country = 'country.name'),
--- a/R/readworldsteel.R
+++ b/R/readworldsteel.R
@@ -56,8 +56,9 @@ readworldsteel <- function(subtype = 'detailed') {
           read_ods(path = file_path, sheet = sheet, na = '...') %>%
             as_tibble() %>%
             mutate(name = sheet) %>%
-            pivot_longer(c(-'country', -'name'), names_to = 'year') %>%
-            mutate(!!sym("year") := as.integer(gsub("^X", "", !!sym("year"))))
+            pivot_longer(c(-'country', -'name'), names_to = 'year',
+                         names_transform = list(year = function(x) {
+                           as.integer(sub('^X', '', x)) }))
         }) %>%
         bind_rows() %>%
         add_countrycode_(origin = c(country = 'country.name'),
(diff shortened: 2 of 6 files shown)


R/readworldsteel.R
```diff
@@ -56,7 +56,8 @@ readworldsteel <- function(subtype = 'detailed') {
           read_ods(path = file_path, sheet = sheet, na = '...') %>%
             as_tibble() %>%
             mutate(name = sheet) %>%
-            pivot_longer(c(-'country', -'name'), names_to = 'year', names_transform = list(year = as.integer))
+            pivot_longer(c(-'country', -'name'), names_to = 'year') %>%
+            mutate(!!sym("year") := as.integer(gsub("^X", "", !!sym("year"))))
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 60: Are you not too young to have done a lot of Pascal programming? :)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
Do you know since when this failed?

In the spirit of more functional programming, I would have gone with 
```
pivot_longer(c(-'country', -'name'), names_to = 'year',
                         names_transform = list(year = function(x) {
                           as.integer(sub('^X', '', x)) }))
```
but you do you.

R/readworldsteel.R
fbenke-pik: Happy to comply

R/calcIEA_EVOutlook.R
```diff
@@ -51,6 +51,8 @@ calcIEA_EVOutlook <- function() {
   # set 0s in other CHA countries than China to approximate CHA as China
   x[c("HKG", "MAC", "TWN"),,] <- 0
 
+  x <- add_dimension(x, dim = 3.1, add = "scenario", nm = "historical")
```
fbenke-pik, line 54: Sorry, sneaked in another minor fix.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> > Do you know since when this failed?
> 
> I am not sure. My theory is that it was hidden behind caching of calcHistorical until yesterday, when I made changes in that calc function, causing recalculations that revealed this problem.

OK, that makes sense.  I was under the erroneous impression that `madrat` also tracked code changes in upstream functions.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> In theory, it should.

Is that a positive or a normative statement? :p

The code only checks for [readSource/calcOutput](https://github.com/pik-piam/madrat/blob/4f0688ddf39907d03d57561fa5eba3d5027d061f/R/getMadratGraph.R#L39) and [tool](https://github.com/pik-piam/madrat/blob/4f0688ddf39907d03d57561fa5eba3d5027d061f/R/getMadratGraph.R#L54) functions.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Are we talking about the same thing?

There was a change in the (upstream) `readODS::read_ods()` function, which `madrat` did not track (and does not claim to track).  So it used the stale cache instead of running the code again, which would have resulted in an error.  `Steel_Statistical_Yearbook_combined.ods` has not been changed since February 2022.

> What I have learnt from asking around over the years (but not validated by looking at the code): if you only changed the underlying file `./Steel_Statistical_Yearbook_combined.ods` , but not `readworldsteel` itself, the caching would not pick it up. 

The `madrat` documentation says otherwise:
> Function which creates a unique fingerprint for a madrat function based on the code of the function itself, other madrat functions which are called by this function and of all source folders involved in the process.

[`madrat:::fingerprint()`](https://github.com/pik-piam/madrat/blob/4f0688ddf39907d03d57561fa5eba3d5027d061f/R/fingerprint.R#L3-L5)

And it at least claims that changes to the source directory do affect the caching:
```
library(mrremind)

x <- madrat:::fingerprint('calcHistorical', TRUE)

grep('steel', names(attr(x, 'details')), ignore.case = TRUE, value = TRUE)
[1] "mrremind:::calcSteelStock"                      
[2] "mrremind:::readworldsteel"                      
[3] "/p/projects/rd3mod/inputdata/sources/worldsteel"
```
But the hashes are calculated based on file names and the first 300 bytes of the files, which might work on binaries, or might not.

> But this does not explain why a change in `calcHistorical` caused recalculation of `readworldsteel`, revealing the bug.

Three theories:
- Did you use an empty cache directory "`falks-cache`"?
- There might have been changes to the `worldsteel` source directory, like a deleted file, which would have changed the fingerprint.
- Different lunar phase.

fbenke-pik:
Thanks for pointing out that file changes do indeed affect the caching, wasn't aware. 
I think I did not use an empty cache directory, but I wouldn't bet on it. But for me the possible explanations are good enough for now.

PR #445, opened 2023-11-08
move calc functions from calcHistorical to appropriate retrieve function

fbenke-pik:
This PR cleans up calcHistorical and moves calc functions directly to the retrieve function fullVALIDATIONREMIND. 

(Inputdata generation using the updated retrieve function succeeded)


R/calcBP.R
```diff
@@ -143,9 +143,9 @@ calcBP <- function() {
   weights[,,"US$2005/GJ", pmatch = T] <- 1
 
   return(list(
-    x = x, 
+    x = x,
     weight = weights,
-    mixed_aggregation = T, 
+    mixed_aggregation = T,
```
LaviniaBaumstark, line 148: TRUE, instead of T ?

LaviniaBaumstark (approved):
Thanks for this nice cleaning!

R/calcBP.R
fbenke-pik: Correct, no idea why this wasn't raised by the linter (probably because, technically I did not change the file, just some whitespaces and line endings were changed when saving it). I will take care of getting these functions linter-conform the next time I actually touch them.

PR #453, opened 2023-12-07
allow filtering for years when creating historical.mif

fbenke-pik:
This PR makes it possible to restrict the years when generating the historical.mif, so we don't need to get all the years starting from the 1700s


R/fullREMIND.R
```diff
@@ -15,7 +13,7 @@
 #' fullREMIND()
 #' }
 #'
-fullREMIND <- function(rev = 0) {
```
LaviniaBaumstark, line 18: did this rev argument has any functionality in retrieveData, which we are only not using?

R/fullVALIDATIONREMIND.R
```diff
@@ -3,16 +3,17 @@
...
+#' @author David Klein, Falk Benke
 #' @seealso
 #' \code{\link{fullREMIND}},\code{\link{readSource}},\code{\link{getCalculations}},\code{\link{calcOutput}}
 #' @examples
 #' \dontrun{
 #' fullVALIDATIONREMIND()
 #' }
 #'
-fullVALIDATIONREMIND <- function(rev = 0) {
+fullVALIDATIONREMIND <- function(years = NULL) {
```
LaviniaBaumstark, line 16: same here regradin `rev`

LaviniaBaumstark (commented):
please, solve the conflicts

LaviniaBaumstark (approved):

PR #465, opened 2024-01-26
remove years argument from fullVALIDATIONREMIND to not break REMIND

fbenke-pik:
This should fix https://github.com/pik-piam/madrat/issues/197


LaviniaBaumstark (approved):
thanks

PR #469, opened 2024-02-07
read in latest IEA data as reference data

fbenke-pik:
Add includes structures to read out latest IEA data to be included in historical mif without updating IEA data for input data generation.

Depends on https://github.com/pik-piam/mrcommons/pull/107


R/calcFE.R
```diff
@@ -28,7 +27,7 @@ calcFE <- function(source = "IEA", scenario_proj = "SSP2") {
     colnames(map) <- gsub("io", "names_in", colnames(map))
 
     # Give description
-    descript <- "IEA Final Energy Data based on 2017 version of IEA Energy Balances"
+    descript <- "IEA Final Energy Data based on 2022 version of IEA Energy Balances"
```
LaviniaBaumstark, line 30: this is only true for *_latest IEA data, right?

R/calcIO.R
```diff
@@ -94,29 +104,31 @@ calcIO <- function(subtype = c("input", "output", "output_biomass", "trade",
...
           as_tibble() %>%
-          select(product = 'iea_product', flow = 'iea_flows'),
+          select(product = "iea_product", flow = "iea_flows"),
 
-        c('product', 'flow')
+        c("product", "flow")
       ) %>%
-      unite('text', c('product', 'flow'), sep = '.') %>%
-      pull('text')
+      unite("text", c("product", "flow"), sep = ".") %>%
```
LaviniaBaumstark, line 117: thanks for all cleaning!

R/calcIO.R
```diff
@@ -225,5 +237,5 @@ calcIO <- function(subtype = c("input", "output", "output_biomass", "trade",
   }
 
   return(list(x = reminditems, weight = NULL, unit = "EJ",
-              description = "IEA SE Output Data based on 2017 edition of IEA World Energy Balances"))
+              description = "IEA SE Output Data based on 2022 edition of IEA World Energy Balances"))
```
LaviniaBaumstark, line 240: would be different when we will update the latest version

R/calcFE.R
fbenke-pik: no, it is true for the current default, i.e the description is accurate for inputdata generation. latest would be 2023. i decided not to make it conditional (to avoid another if-clause) in all places where we add a description and accept that the description is inaccurate when using it solely for generating the historical.mif.

R/calcIO.R
fbenke-pik: see me comment above. this is accurate for the default, aka the inputdata generation. when we update to latest version, we will have to update here, too (and in many other places)

LaviniaBaumstark: o.k., hope we will remember

LaviniaBaumstark (approved):

PR #472, opened 2024-02-12
Check for version incement for PRs

fbenke-pik:


pascal-sauer (commented):
On second thought, setting up the whole dependencies again just to run a tiny command seems wasteful. Could we integrate this into our main github action, and only run it if the action was triggered by a PR (and not a push)?

PR #478, opened 2024-02-27
refactor reading in UNFCCC data

fbenke-pik:
This PR simplifies logic in convertUNFCCC by using madrat functions and unifies handling of 0s vs NAs in calcUNFCCC:
all countries in regions "EUR", "REF", "NEU", "CAZ" are set to 0s if they have NAs so regional aggregation yields results.

Before, this was limited only to data read in from the source, now it also includes data calculated on top of read in data.

This fixes the problem that
`Emi|CO2|Industrial Processes` and `Emi|CO2|Energy|Demand|Industry` are available for EU27 and CAZ, but `Emi|CO2|Industry ` is not.

### DIFF
--- a/R/calcUNFCCC.R
+++ b/R/calcUNFCCC.R
@@ -44,8 +44,6 @@ calcUNFCCC <- function() {
     as.magpie() %>%
     toolCountryFill(fill = NA, verbosity = 2)
 
-  mappedVariables <- getNames(x)
-
   # aggregate pollutants ----
 
   x <- add_columns(x, "Emi|CH4 (Mt CH4/yr)", dim = 3.1)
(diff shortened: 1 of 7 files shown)


R/calcUNFCCC.R
```diff
@@ -239,11 +236,14 @@ calcUNFCCC <- function() {
 
   # return results ----
 
-  # set 0 to NA in calculated variables
-  calculatedVariabes <- setdiff(getNames(x), mappedVariables)
-  tmp <- x[, , calculatedVariabes]
-  tmp[tmp == 0] <- NA
-  x[, , calculatedVariabes] <- tmp
+  # fill countries of selected regions with 0 to allow for regional aggregation
+  regions.fill <- c("EUR", "REF", "NEU", "CAZ")
```
robertpietzcker, line 240: can you please also add "EU27" to this list?

robertpietzcker (approved):
I think the logic is good - can't say much about the actual R code changes :-)

R/calcUNFCCC.R
fbenke-pik: it should be implicitly included by EUR, as EU27 is a subset of EUR

PR #485, opened 2024-03-15
add calc function for waste energy use shares

fbenke-pik:


Renato-Rodrigues (changes requested):
specific regional mapping aggregations should not be used inside of calc functions to keep support with multiple regional resolutions.

R/calcWasteEnergyUseShares.R
```diff
@@ -1,64 +1,90 @@
...
   return(list(
-    x = out,
-    weight = weight,
+    x = numerator,
+    weight = NULL,
     unit = "%",
-    mixed_aggregation = TRUE,
+    aggregationFunction = .calcShares,
+    aggregationArguments = list(denominator = denominator),
+    min = 0, max = 1,
```
Renato-Rodrigues, line 87: Are the min and max parameters necessary?

Renato-Rodrigues (commented):
Thank you for the changes added!
I don't have any additional suggestions, besides doing a last check for a given region to make sure the values makes sense on H12 and EU21 regional aggregations.

R/calcWasteEnergyUseShares.R
fbenke-pik: No, it is just an optional check that makes sense to me, as these are shares and should always be between 0 and 1.

PR #487, opened 2024-03-21
add WEO 2023 free data to historical.mif

fbenke-pik:
- this PR adds the 2023 free data to historical.mif
- renames functions for consistency with future data updates (read/convert/calc: IEA_WEO_2021 -> IEA_WorldEnergyOutlook)
- refactors calcOutput function to use a custom aggregation function 

- fixes a bug in convert function:
  - before: convert function disaggregates EU to countries according to a weight, but then sets GIB to 0
  - values should be slightly higher after this bugfix 
  - this also affects calcFeDemandIndustry cc @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q


R/calcFeDemandIndustry.R
```diff
@@ -964,7 +964,7 @@ calcFeDemandIndustry <- function(use_ODYM_RECC = FALSE) {
         industry_subsectors_en %>%
           filter(2010 == .data$year),
 
-        readSource(type = "IEA_WEO_2021", subtype = "region",
```
orichters, line 967: not worried about backwards compatibility? But you certainly much more than I do about the whole process.

R/readIEA_WorldEnergyOutlook.R
```diff
@@ -0,0 +1,100 @@
...
+        file = "2023/free/WEO2023_AnnexA_Free_Dataset_Regions.csv",
+        sep = ","
+      ) %>% rename_all(tolower),
+      read.csv2(
+        file = "2023/free/WEO2023_AnnexA_Free_Dataset_World.csv",
+        sep = ","
+      ) %>% rename_all(tolower)
+    ) %>%
+      mutate(
+        "value" = ifelse(.data$unit == "PJ", as.numeric(.data$value) / 1000, as.numeric(.data$value)),
```
orichters, line 84: while correct, I don't understand for which variables in the mapping this conversion is required

orichters (approved):

R/calcFeDemandIndustry.R
fbenke-pik: Don't see any issues with backwards compatibility.
- `madrat::getDependencies("readIEA_WEO_2021", direction = "dout")` tells you, this function is only used here and in the corresponding calc function
- running older versions of mrremind won't crash, because the original folder structure is left on the cluster:
`/p/projects/rd3mod/inputdata/sources/IEA_WEO_2021

R/readIEA_WorldEnergyOutlook.R
fbenke-pik: The source sometimes uses PJ as unit, e.g. for Energy-Hydrogen-Final demand

LaviniaBaumstark (approved):

PR #497, opened 2024-05-17
update EMBER electricity data to 2024

pweigmann:
This data update doesn't require any change besides updating the source file. As last years data was already put in a folder "2024" even though it was from 2023, it will be moved to a folder 2023 and the new file will replace it.

To understand the changes which the data update brings (besides adding data for the year 2023), a validation report was created and is available here:

https://cloud.pik-potsdam.de/index.php/f/20716881

It shows the deviation of the 2024 EMBER release to the 2023 one, paints yellow deviations above 10%, and red above 25%.

There might be some relevant changes, also in older periods, especially for gas capacities. @RahelMA @robertpietzcker

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '35343680'
+ValidationKey: '35370660'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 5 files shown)


fbenke-pik (approved):

robertpietzcker:
Thanks for the heads-up. 
I think it looks good. For the EU, the new values are much closer to what the WEO reported in those days, so it is likely an improvement.

PR #500, opened 2024-05-23
Refactor according to mrdrivers update

johanneskoch94:
Move readPWT and readOECD over from mrdrivers, where they aren't used any longer.
Adjust calls to readIMF.
Drop old GDP sources from calcHistorical.


johanneskoch94:
Also, when I ran the input data, my version of mrremind didn't include your latest commits Michaja, in case that plays a role.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
How does this branch relate to commit `ace22a8` used for `rev6.75jokoch_2005PPP`?

johanneskoch94:
Is ace22a8 the commit of mrremind used in the input-data creation? If so, it got rewritten when I rebased this branch in preparation of the merge to develop... Different hash, same code. (I realize now that this breaks the link between the current input data version and the r-package, but the commits were a mess, and I was relying on the fact that we will rerun the input data once all packages are merged...)

LaviniaBaumstark (approved):

PR #503, opened 2024-05-31
Fix warnings

fbenke-pik:
This fixes two obnoxious warnings in inputdata generation
1. This type of warnings is spamming the log and has its origin in `readGGDC10`: `Expecting numeric in O7550 / R7550C15: got 'incl in gov. services'`
2. Apparently madrat does static code analysis when loading it and throws a misleading warning when you call toolGetMapping with parameters, as done previously in calcIO. Therefore, I rewrote it to a more explicit (but more redundant) version that madrat should understand (cc @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q )

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '35601664'
+ValidationKey: '35634082'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'mrremind: MadRat REMIND Input Data Package'
-version: 0.179.2
-date-released: '2024-05-24'
+version: 0.179.3
+date-released: '2024-05-31'
 abstract: The mrremind packages contains data preprocessing for the REMIND model.
 authors:
 - family-names: Baumstark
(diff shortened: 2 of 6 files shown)


LaviniaBaumstark (approved):

fbenke-pik:
> Say whaaat? madrat happens to magically do static code analysis, that nobody put in there? 

Loading madrat previously produced this warning
`1: In .warnNotfound(getMappings) : Mapping in mrremind:::calcIO not found!`

The corresponding code can be found here: https://github.com/pik-piam/madrat/blob/master/R/getCode.R

`getCode` is called by `getDependencyGraph`.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
1. When would one ever see those, instead of the cheery `There were 50 or more warnings (use warnings() to see the first 50)` when working with `madrat`? :confused: 
   Nonetheless, the warning did exist, and I feel silly for my comment.  On the up side, one less warning.
2. Looking at `madrat::getCode()`, I think this change sets a bad precedent.
   In my opinion it is not helpful to try to reason about code by evaluating it (which by the way is the opposite of _static_ code analysis) separate from its execution environment.  All instances of `.warnNotfound(getMappings)` I found (four times `mredgebuildings`, and `mrremind::calcIO()`) are false negatives where the mapping name is composited.
   This encourages repetitive, therefore redundant, therefore error-prone code.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
:point_right: pik-piam/madrat#209

PR #509, opened 2024-06-13
Drop dependency on assertthat and refactor calcTheil

johanneskoch94:


johanneskoch94:
Hey, I tagged you Björn because you're listed as author... but perhaps I should tag someone else? In any case, I just refactored the function, while I was dropping the assertthat dependency :)

bs538 (approved):
Thanks. While I generally prefer assert_that() over stopifnot(), reducing dependencies makes sense. And of course good to clean up.

johanneskoch94:
Yeah, assertthat is a bit nicer that stopifnot, but since it really is the only appearance in the package (and in all our packages for that matter) I thought it was cleaner to use stopifnot... Thanks for the review!

PR #522, opened 2024-07-15
fix errors with inputdata generation occurring under R 4.3.2

fbenke-pik:
This fixes enconding issues in readGCPT that cause errors under R 4.3.2. 
While we're at it, i fixed some of the more "severe" linter warnings and applied autoformat.


LaviniaBaumstark (approved):
nice, thanks

PR #529, opened 2024-08-01
update IEA WEO reference data to 2023 data

fbenke-pik:
This PR replaces reading in WEO 2021 complete data and 2023 free data by 2023 complete data.

Changes:
- Updated the variable mapping to REMIND, as some of the terminology has changed in the source data
- As of now, we no longer have the variables `Emi|CO2|CDR|DACCS`, `Emi|CO2|CDR|BECCS (Mt CO2/yr)`, `Carbon Management|Storage (Mt CO2/yr)`  (they could be brought back once the mappings are updated by sector experts)
- The scenario  `Sustainable Development Scenario`  is no longer provided with the new data

Data comparison: https://cloud.pik-potsdam.de/index.php/f/21536125


R/calcFeDemandIndustry.R
```diff
@@ -951,20 +951,21 @@ calcFeDemandIndustry <- function(use_ODYM_RECC = FALSE) {
     mutate(pf = sub("^feelwlth_", "feel_", .data$pf))
 
   ### extend to 2020 using WEO 2021 data ----
-  IEA_WEO_2021_ref_year <- 2020L
+  # valid possibilities in 2023 data: 2021 and 2022
+  IEA_WEO_2023_ref_year <- 2020L
   last_industry_subsectors_en_year <- max(industry_subsectors_en$year)
-  if (IEA_WEO_2021_ref_year > last_industry_subsectors_en_year) {
```
fbenke-pik, line 956: @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q WEO 2023 no longer has 2020 data, but 2021 and 2022. However, this section was never executed, because the condition `IEA_WEO_2021_ref_year > last_industry_subsectors_en_year` was not met.

If you set the ref year to 2021 or 2022, the section will crash due to the assert statement at the end. But this is also true for the 2021 data, so you will have to look into this anyways, if you want this section to be executed.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: I am quite confident that this code was executed before the update to the IEA Energy Balances, which then only reached to 2016.  But since that update, this code is obsolete and self-deactivated as intended.
I suggest deleting this section (and checking the calibration data with old and new WEO data, of course).

LaviniaBaumstark (approved):

PR #530, opened 2024-08-01
introduce pipeline subtype, update IEA CCUS and GEM

pweigmann:
**PR Update**
I extended this PR to include a more thorough approach of handling near-term pipeline data and more generally create output files for min/max thresholds which can be used with ``piamValidation`` and offer better possibilities to analyze project pipelines.

The current implementation offers the following options:
- use ``exportThresholds(type = "full")`` to assemble threshold variables (e.g.  ``Cap|Electricity|Hydro|min_red``) and the data that was used to derive it (e.g. ``Cap|Electricity|Hydro|announced``) from ``calcProjectPipelines`` and potentially other functions in the future. This data is exported to ``pipelines.mif`` in the madrat output folder.
- use ``exportThresholds(type = "config")`` to export only the thresholds (``min/max_yel/red``) in a way that they can be easily copied to a ``validationConfig`` from ``piamValidation``. The output file is called ``thresholds.mif``.
- in ``calcProjectPipelines``, theses thresholds are calculated using assumptions and expert knowledge

The PR also updates the Global Energy Monitor data to 2024, using the integrated publication "Global Integrated Power Tracker" instead of individual trackers per technology. 

GEM and IEA CCUS data for the historical.mif or used as input-data to REMIND is updated but for now kept separately to the ``pipeline`` formulations (the old subtypes did not change).

**Old PR**
This PR does two things:

1.) update the IEA CCUS data from 2023 to 2024. This results in one more year of historical data via ``calcCCScapacities(subtype = "historical")`` and more recent additions to the projections used as input data (``calcCCScapacities(subtype = "projections")``)

2.) introduce a new subtype ``pipeline`` which doesn't pre-select depending on the ``status`` of a project in the ``readSource``-function and then can be used to formulate thresholds as used in the validation tool in the new function ``calcProjectPipelines(subtype = CCS)``. The idea would be to supply data from other similar project pipelines to this function as well in later work (hydro, nuclear, wind capacities etc.).

@fbenke-pik @amerfort what do you think about this approach? ``projections`` and ``pipeline`` are quite similar but Gunnar and the Carbon Management team seem to have different preferences for assumptions and the data is used in different ways.

@amerfort regarding the data update, I didn't generate new input-data or did a test run of REMIND with the new numbers - do you think the update could have strong effects?


pweigmann:
Here is a preview of the new data: [ccs.csv](https://github.com/user-attachments/files/16458968/ccs.csv)

R/calcProjectPipelines.R
```diff
@@ -0,0 +1,66 @@
+#' calc Project Pipelines
```
fbenke-pik, line 1: Please make sure to note where this will be called. When I clean up old code, I usually run sth like
`madrat::getDependencies("calcProjectPipelines", direction = "dout")`  and if I find no results, I might just delete the code.

R/calcProjectPipelines.R
```diff
@@ -0,0 +1,66 @@
...
+#' Calculate the expected near-term deployment of technologies
+#'
+#' @author Pascal Weigmann
+#'
+#' @param subtype choose technology, either `CCS`, `Nuclear`
+#'
+#' @export
+calcProjectPipelines <- function(subtype) {
+  if (subtype == "CCS") {
+    x <- readSource("IEA_CCUS", subtype = "pipeline")
```
fbenke-pik, line 12: This section is redundant to`calcCCScapacity` capacity. I assume you looked into removing the redundancy, but did not find an easy way to do so? If that is the case, then it is fine by me.

R/calcProjectPipelines.R
fbenke-pik: Edit: just read you comment concerning this. I am fine with the redundancy in this case. It gives you flexibility, should requirements by the different teams diverge in the future.

fbenke-pik (approved):

R/calcProjectPipelines.R
fbenke-pik: Edit: just read you comment concerning this. I am fine with the redundancy in this case. It gives you flexibility, should requirements by the different teams diverge in the future.

R/calcProjectPipelines.R
pweigmann: Will do, so far it is not called anywhere, but I assume sooner or later it will be used either in ``fullValidation`` or some similar new function.

R/calcProjectPipelines.R
pweigmann: Agreed, it is not great, ideally the two approaches will be merged in the future.

amerfort:
For the REMIND input, we can also work with "Gunnars logic" as it has all information that we need too. So no need to keep the redundancy from our side. However, I'm not experienced with how the function during input data generation creates the .cs4r input file that REMIND uses then. I just have to make sure it has the standard REMIND-readable format, i.e. 
`period`, `region`, `variable`, `value` and not the time steps as columns as in your csv.Would that be possible? You can have a look at the current input data file here: `/p/projects/remind/modeltests/remind/core/input/pm_boundCapCCS.cs4r`
(Or basically in any current REMIND folder)

pweigmann:
> For the REMIND input, we can also work with "Gunnars logic" as it has all information that we need too. So no need to keep the redundancy from our side. However, I'm not experienced with how the function during input data generation creates the .cs4r input file that REMIND uses then. I just have to make sure it has the standard REMIND-readable format, i.e. `period`, `region`, `variable`, `value` and not the time steps as columns as in your csv.Would that be possible? You can have a look at the current input data file here: `/p/projects/remind/modeltests/remind/core/input/pm_boundCapCCS.cs4r` (Or basically in any current REMIND folder)

Perfect, that will make things easier. Will make sure it has the same format, it is created [here](https://github.com/pik-piam/mrremind/blob/c53b4c1ba8d637a6bdc11b97328f86e2b2efc780/R/fullREMIND.R#L131), directly via ``calcOutput("CCScapacity", subtype = "projections")``.

pweigmann:
@fbenke-pik @PhilippVerpoort I updated this PR to include a more general approach on handling project pipelines and near-term thresholds in mrremind. Let me know what you think!

PR #533, opened 2024-09-02
turn 'exportThresholds' into a full function

fbenke-pik:


pweigmann:
Can be closed after merging https://github.com/pik-piam/mrremind/pull/548

PR #536, opened 2024-09-04
update IEA Energy Balances

fbenke-pik:
Update reference data to July 2024 data. This gives us an additional year (2022).
See comparison to July 2023 here: https://cloud.pik-potsdam.de/index.php/f/22077519

### DIFF
--- a/R/calcTrade.R
+++ b/R/calcTrade.R
@@ -23,6 +23,6 @@ calcTrade <- function() {
 
   return(list(
     x = x, weight = NULL, unit = "EJ",
-    description = "IEA Final Energy Data based on 2023 version of IEA Energy Balances"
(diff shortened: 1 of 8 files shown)


R/calcTrade.R
```diff
@@ -23,6 +23,6 @@ calcTrade <- function() {
 
   return(list(
     x = x, weight = NULL, unit = "EJ",
-    description = "IEA Final Energy Data based on 2023 version of IEA Energy Balances"
+    description = "IEA Final Energy Data based on 2024 version of IEA Energy Balances"
```
LaviniaBaumstark, line 26: does this mean an update for teh input data or only for the validaiton data?

fbenke-pik: calcTrade is used to generate historical.mif only

LaviniaBaumstark (approved):

PR #537, opened 2024-09-05
filter IEA ETP production data using a custom aggregation function

fbenke-pik:
- This PR uses a custom aggregation function to read in GLO data directly, rather than disaggregating and aggregating.
- Besides, the regional disaggregations of the regions OECD and Non-OECD were dropped, because the disaggregated data relies heavily on the chosen weight and is not usable for comparison. 
- This results in some regions having less variables (because the values before were unreliable) and CHA having lover values than before (because it now equals CHN, with other CHA regions set to 0).
- For details, see: https://cloud.pik-potsdam.de/index.php/f/22298897
- This does not affect industry inputdata generation, as the corresponding functions `calcFeDemandIndustry` and `calcSteel_Projections`  only use readIEA_ETP without the convert function and this portion has not been changed (only enhanced by World data)


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (commented):
Why?

fbenke-pik:
Not really a discussion. It is basically the madrat-way. Condsider it FYI

LaviniaBaumstark (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #560, opened 2024-09-27
fix lintr warnings and other format problems

pweigmann:


pweigmann:
It seems git doesn't handle it well to change only the capitalization of a filename...

PR #564, opened 2024-10-04
update of nuclear thresholds

pweigmann:
fix name of IAEA PRIS source
add inactive reactors in Japan
use PRIS and Ember for 2020 nuclear reference capacities

See: https://github.com/pik-piam/mrremind/discussions/540


RahelMA (approved):
Thank you, Pascal!

PR #574, opened 2024-10-11
explicitly monitor GDPuc::convertGDP for madrat caching

### LINKED ISSUE #573: GDPuc is not registering with madrat, therefore does not affect caching
Which is bad, since pik-piam/GDPuc#37 is transparent to madrat, which happily loads everything calculated using faulty dollar conversion factors from the cache.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
fix #573

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '38731616'
+ValidationKey: '38753559'
 AcceptedWarnings:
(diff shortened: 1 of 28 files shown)


fbenke-pik:
Can be closed in favour of https://github.com/pik-piam/mrremind/pull/575

PR #575, opened 2024-10-11
replace unit conversion functions with madrat tool wrappers

fbenke-pik:


DESCRIPTION
johanneskoch94, line 1: Does GDPuc now have to be listed under Depends? I don't know..

R/EDGE-Industry.R
johanneskoch94, line 1: And calling a tool function with `::` doesn't hinder the caching right?

johanneskoch94 (commented):
Apart from the above comments, it looks good. (Some indentation was also messed up, but that can wait.)

DESCRIPTION
fbenke-pik: As far as I can tell, no

R/EDGE-Industry.R
fbenke-pik: It should not.

johanneskoch94 (approved):

PR #576, opened 2024-10-11
Adjust to new EDGE buildings output structure

ricardarosemann:
Adjust the mrremind code to the changed output structure of EDGE buildings:
- Require mrcommons version with updated structuremapping
- Remove adjustments for Navigate and Campaigners scenarios
- read all data from one file ```EDGE_buildings.csv``` instead of two separate files ```EDGE_buildings_energy.csv``` and ```EDGE_buildings_floorspace```
- In the EDGE buildings data, the RCP scenario (if existent) is now assumed to be specified as a suffix to the scenario, with any commas replaced by "_", e.g. "SSP1_rcp4_5"
- The output of ```readEdgeBuildings``` and depending on the subtype, also of ```calcFeDemandBuildings``` contains an "rcp" column with the RCP scenario. If no RCP scenario was given, the value of the "rcp" column now consistently is "none".


robinhasse (approved):
Looks good, thank you @ricardarosemann!

PR #582, opened 2024-10-28
fix industry demand to SSP2 until fixing_year

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
See remindmodel/development_issues#384.
- industry UE and FE demand are fixed to SSP2 values until fixing_year
- controlled via `/p/projects/rd3mod/inputdata/sources/industry_subsectors_specific/fixing_year.csv`
- currently set to 2025 for all but `SSP1` (where it is 2028)


laurinks:
Thanks, @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q! Do I get it right that `fixing_year` is precisely the additional flexibility you meant [here](https://github.com/remindmodel/development_issues/issues/384#issuecomment-2439918231) when writing _"I will adjust the code some more"_ ?
I think, I don't fully get yet how the two options (fixing_year = 2025 (default) and fixing_year = 2028 (ScenarioMIP specific setting)) will become available for REMIND runs. Are you generating separate input data trajectories for the two options, and then we simply choose when making a calibration run? Thanks for clarifying. And sorry if that is a dump question (I have not been looking into input data and calibration until now).
@JakobBD : What would that mean time-wise?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I think, I don't fully get yet how the two options (fixing_year = 2025 (default) and fixing_year = 2028 (ScenarioMIP specific setting)) will become available for REMIND runs. Are you generating separate input data trajectories for the two options, and then we simply choose when making a calibration run? Thanks for clarifying.

If you want a scenario to be fixed to SSP2 longer than until 2025, you have to say so in `/p/projects/rd3mod/inputdata/sources/industry_subsectors_specific/fixing_year.csv`.  (Actually, that is a per-region setting, because it meant less code changes.  For all regions, just put `NA` there.)  That happens all on the input data side and requires calibration.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I'm currently in the process of moving some of these files to mrindustry. That will be a mess in merging. Could we please wait with this PR until the migration is finished?

No.  I was totally planning on throwing you under the bus with the pull request and break both of you package branches!

JakobBD:
> @JakobBD : What would that mean time-wise?

Later today, if all goes well (it never does)

> No. I was totally planning on throwing you under the bus with the pull request and break both of you package branches!

I expected no less.

JakobBD:
Re-opened in pik-piam/mrindustry#2. 

Only comes into effect there once pik-piam/mrremind#581 is merged, but that should be later today or tomorrow, according to @fbenke-pik. 

@laurinks is that a timeframe you can live with?

laurinks:
@JakobBD: I am doing tests currently with input data version `rev7.1-i2` created by @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, which I can do independently of this merge. Thus, it is not urgent for me. Anyway, I am off next week, so things get merged next week and become available the week after, that is good for me. I suppose it will be the same for @lecfab who is at IAMC next week.

PR #591, opened 2024-11-15
update input file to REMIND_11Regi

fbenke-pik:
This PR adjusts the capacity factor of nuclear in Japan in 2025 and 2030 downwards, as the restarting of reactors happens much slower than initially expected.


robertpietzcker (approved):

robertpietzcker:
This PR adjusts the capacity factor of nuclear in Japan in 2025 and 2030 downwards, as the restarting of reactors happens much slower than initially expected.

PR #592, opened 2024-11-15
update regionalized PV investment costs

fbenke-pik:
Robert Pietzcker: Partial update of regionalized PV investment costs for 2020, as previous data was based on a 2018 publication.

### DIFF
--- a/R/calcDiffInvestCosts.R
+++ b/R/calcDiffInvestCosts.R
@@ -52,8 +52,8 @@ calcDiffInvestCosts <- function(subtype) {
     for (y in years) {
       x[
         getRegions(x)[x[, y, "Coal.Steam Coal - SUBCRITICAL"] != 0 &
-          x[, y, "Coal.Steam Coal - SUPERCRITICAL"] != 0 &
-          x[, y, "Coal.Steam Coal - ULTRASUPERCRITICAL"] != 0], y,
+                        x[, y, "Coal.Steam Coal - SUPERCRITICAL"] != 0 &
+                        x[, y, "Coal.Steam Coal - ULTRASUPERCRITICAL"] != 0], y,
         c(
           "Coal.Steam Coal - SUPERCRITICAL",
           "Coal.Steam Coal - ULTRASUPERCRITICAL"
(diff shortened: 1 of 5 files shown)


robertpietzcker (approved):

R/calcDiffInvestCosts.R
```diff
@@ -124,28 +124,33 @@ calcDiffInvestCosts <- function(subtype) {
...
+    x_adj["LAM", , ] <- 850
     # MEA: IRENA states utility-scale prices in the range of ~1250$/kW for MEA
     # Recent auctions in individual countries indicate lower prices,
     # but it is unlikely that all the countries will immediately achieve
     # the lowest prices realized in some auctions
-    x_adj["MEA", , ] <- 850
+    x_adj["MEA", , ] <- 800
     # REF: Very different costs in IRENA (2300) and REN21 (1300) -
     # but unlikely to have higher capital costs than severly space-constrained Japan
     x_adj["REF", , ] <- 1400
```
lecfab, line 150: The comment says that for capital cost REF < JPN, but the new values don't match that (1400 vs 1100).
Which one is correct:
- REF may have higher costs than JPN?
- REF costs are below 1100?

fbenke-pik: @robertpietzcker

robertpietzcker: Good catch, Fabrice : the comment on REF was written with respect to the old values from 2018, and the change I now made to JPN was based on newer IEA PVPS data for JPN. 

I know little about REF PV costs. 

I see two options: 
a) leaving REF as is, and adding a "much" in the sentence, as in "unlikely to have much higher capital costs"
b) reducing REF to 1100. 

any thoughts on this, @lecfab ?

lecfab: I'm also very ignorant on REF PV costs.
There is this [arxived paper](https://arxiv.org/pdf/2312.07240) (trying to assess whether russian pv can go without subsidies) that shows the following costs. Not sure what they exactly encompass. As of today, 1RUB ≃ 0.01$. The capital cost would range between 600 and 1200 $/kW.

image

A few examples in Kazakhstan seem to get capital costs between [500 ](https://mergerscorp.com/property/20mw-kazakh-pv-photovoltaic-power-plant-gov-ppa/) and [850](https://renewablemarketwatch.com/country-reports/renewable-projects-database/solar-pv-projects/kazakhstan-35-mwp-solar-photovoltaic-pv-power-plant-with-building-permit-project-profile-single-user-detail) $/kW. I am not sure these sources are reliable, and I don't know if everyone means the same thing with capital cost.

Maybe option b) is better for consistency? is there someone who has general gut feelings for REF?

robertpietzcker: I can't think of anyone else to ask. 
Given the numbers you found, I think going for b) sounds reasonable

@fbenke-pik could you do that (changing `x_adj["REF", , ] <- 1400` to `x_adj["REF", , ] <- 1100`)? 
I guess it needs a new PR, or can we simply include it in this one?

lecfab: (done in this commit: https://github.com/pik-piam/mrremind/commit/6e894af7a61983e105e6bcc868e9cc6d3ef0b7d7)

PR #600, opened 2024-12-02
remove Stationary from calcFeDemand

fbenke-pik:


fbenke-pik:
Closed in favour of https://github.com/pik-piam/mrremind/pull/604

PR #601, opened 2024-12-02
fix bug in calcCCScapacity

fbenke-pik:


RahelMA (commented):
I'm uncertain here. From what I understand, I would say it influences the bound calculation 

> "CCS capacity derived from IEA CCUS project database"

The ideal case would be to have unchanged IEA CCUS historical data until 2023 and in addition, the estimates on bounds (as additional variable). 

@pweigmann could you please help out here?

pweigmann:
> I'm uncertain here. From what I understand, I would say it influences the bound calculation
> 
> > "CCS capacity derived from IEA CCUS project database"
> 
> The ideal case would be to have unchanged IEA CCUS historical data until 2023 and in addition, the estimates on bounds (as additional variable).
> 
> @pweigmann could you please help out here?

This is the right approach and should not impact the bounds, as they are using the subtype ``pipeline``. This is fixing an oversight that I probably introduced when reworking the function.

pweigmann (approved):

PR #602, opened 2024-12-03
Remove luscale dependency

fbenke-pik:
This PR removes some warnings in inputdata generation. 
Removes outdated luscale functions and replaces them with toolAggregate.


LaviniaBaumstark (approved):
thanks

PR #620, opened 2025-01-28
Move calcRatioPPP2MER over from mrdrivers 

johanneskoch94:
- Move calcRatioPPP2MER over from mrdrivers and fix documentation
- Bump version

The line `Roxygen: list(markdown = TRUE)` in the DESCRIPTION file enables the inline code in the documentation and inline linking to function and packages, as is for example the roxygen comments of calcCapital and calcRatioPPP2MER. The .Rd files of all other functions are touched automatically then by roxygen.


fbenke-pik:
Thanks for the improvement, Johannes!

fbenke-pik (approved):

PR #622, opened 2025-01-29
New sources for non-CO2 emission references

gabriel-abrahao:
This PR introduces `calcEmiMacCEDS` for generating CEDS2024-based non-CO2 emissions and adds an option to `calcMACCbaseN2O` to read baseline transport and acid production N2O emission trajectories from Harmsen et al. 2023 instead of the old Lucas et al. 2007 IMAGE baseline. 

All files created are new, not affecting existing files.


merfort (approved):
Thanks, looks good.

nicobauer:
Do we have some test runs?

gabriel-abrahao:
@nicobauer we do, but I'm putting them in the REMIND PR that comes next. However, good point, for documentation's sake here's some comparisons of the new N2O baselines for 2020 and 2050. Mostly upward changes, that is, we'll get higher N2O emissions from these sectors as they are completely exogenous. Notably, USA transport and adipic acid emissions are lower in the new basline.

dummy

And the new data that will be used to "calibrate" the econometric estimates and rescale the baseline. We currently use EDGAR_2005 here, which will be replace with CEDS_2020. CEDS_2005 is just to there to give a sense of what's the difference in databases vs. in base year. It's a bit of mixed bag. Waste, sewage and landfill emissions are generally higher but not everywhere, and transport and acid N2O is often lower, so we'll get a downward adjustment of the above baselines in REMIND itself.

image
d

PR #625, opened 2025-01-29
Refactor fullRemind in line with new gdp and pop GAMS sets

johanneskoch94:
Refactor fullRemind in line with new gdp and pop GAMS sets.
Add scenario specification for all calcGDP, calcGDPpc and calcPopulation calls, in anticipation of mrdrivers changes.
Clean up NAMESPACE file by removing importFrom statements and adding :: calls for select functions.


R/calcFEdemand.R
```diff
@@ -9,7 +9,7 @@ calcFEdemand <- function() {
 
   # add Navigate and Campaigners scenarios to industry and transport to match buildings scenarios by duplication
   duplicateScens <- "gdp_SSP2EU_NAV_all"
-  feIndustry <- mbind(feIndustry, setItems(feIndustry[, , "gdp_SSP2EU"], 3.1, duplicateScens))
+  feIndustry <- mbind(feIndustry, setItems(feIndustry[, , "gdp_SSP2"], 3.1, duplicateScens))
```
LaviniaBaumstark, line 12: this was a bug as we are now using SSP2 for all other data, thanks for fixing it

PR #630, opened 2025-02-11
Add distribution of Exclusive Economic Zone (EEZ) area 

tabeado:
This PR adds the source "MarineRegionsOrg" which provides data on the size of the exclusive economic zones (EEZ). 

The output is the share in total EEZ area by country/region. It will be used in 33_cdr for distribution of OAE.


fbenke-pik (approved):

fbenke-pik:
Looks good! Don't forget buildLibrary.

robertsalzwedel (approved):
Logic in function seems good! :)

PR #633, opened 2025-02-14
Implement explicit scenario choice across package.

johanneskoch94:
- Introduce a scenario argument to calcCapital, calcFeDemand, calcFeDemandBuildings, calcFloorspace, calcTheil, and calcEmiTarget. The larger picture is the attempt to step away from a default set of scenarios. The choice of scenarios should be explicit. In the case of the input data pipeline, it is made in fullREMIND and then passed on to the different calcOutput functions. Hopefully this will help us control/drop/introduce new scenarios more easily in the future.
- Move SSP2_NAV_all duplication to mrindustry.
- Set NAs in CO2 prices from ExpertGuess to 0.
- Drop unused sources argument in calcEmiTarget.
- Drop unused scenario_proj argument in calcFE.
- General refactoring.

!! Still waiting for my input data test to run through (/p/tmp/jokoch/pre-processing, output at /p/projects/rd3mod/inputdata/output_1.27/rev7.22johannes_test7_62eff8f7_remind) but tagged in the reviewers so that the revision can start.

Connected to this [PR](https://github.com/pik-piam/mrindustry/pull/29) in mrindustry.


R/readStationary.R
```diff
@@ -1,37 +1,166 @@
...
+#' is used to specify the desired scenarios. Unspecified scenarios are copies of the SSP2 scenario.
 #'
-#' @author Antoine Levesque, Robin Hasse
+#' @return magclass object
+#' @order 1
 readStationary <- function() {
+  read.magpie("EDGE_TradMod.cs4r")
+}
 
-  scenarios <- list(SSPs  = paste0("SSP", 1:5),
```
LaviniaBaumstark, line 8: Did you talk about read- and convertStationaryy with @fbenke-pik ?

johanneskoch94: No, ist there a something I should know @fbenke-pik ?

fbenke-pik: it will be removed at some point, but for now we need to keep it.

johanneskoch94:
Side note: 
I see this warning popping up at in multiple calculations. It seems to come from something happening in calcOutput. How do I get rid of it?
```
WARNING:
~ Weight sum is 0, so cannot normalize and will return 0 for some aggregation targets. This changes the total sum of the magpie object! If this is really intended set zeroWeight = "allow", or "setNA" to return NA.
```

R/calcFEShares.R
```diff
@@ -2,28 +2,32 @@
...
-
-  #---- Read Data -----
+  if (! subtype %in% c("ind_coal", "ind_bio", "ind_liq")) {
+    stop("Unknown subtype.")
+  }
 
   edge_buildings <- calcOutput("IOEdgeBuildings", subtype = "output_EDGE_buildings", aggregate = FALSE)
   output <- calcOutput("IO", subtype = "output", aggregate = FALSE)
-  fe_demand <- calcOutput("FEdemand", aggregate = FALSE)[, , "SSP1"]
+  fe_demand <- calcOutput("FEdemand", scenario = "SSP1", aggregate = FALSE)
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 12: Disables the `madrat` cache for this call.  See https://github.com/pik-piam/mrindustry/pull/30#discussion_r1965132521 and https://github.com/pik-piam/mrindustry/pull/30#discussion_r1965141924.  Should be fixed for the benefit of run times and disk space.  And is probably an issue across related changes in other packages.

johanneskoch94: You mean it doesn't use a potential cache, in which the SSP1 scenario could be available right? I don't see a way around this without keeping a "default" set of scenarios, which is exactly what we want to get rid of.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: `calcFEShares()` is called by `fullREMIND()`, and that knows what scenarios are going to be calculated.  Just pass the `scenario` [sic] parameter passed to e.g. `calcFEdemand()` to `calcFEshares()` too.

johanneskoch94: Ok! Thanks, will implement this right away.

robertsalzwedel (approved):
@johanneskoch94 asked me to approve this as a technicality.

R/readStationary.R
```diff
@@ -1,37 +1,168 @@
...
-      dimName = "scenario",
-      dimCode = 3.1
-    )
-  )
+#' @rdname readStationary
+#' @order 2
+#' @param x MAgPIE object returned from readStationary
+#' @param subset A string, or vector of string designating the entries for the scenario dimension. If possible,
+#'  passed to [mrdrivers::calcGDP()]. Unspecified scenarios receive SSP2 data.
+convertStationary <- function(x, subset) {
```
fbenke-pik, line 17: @johanneskoch94 moving read and convert into the same file is a big no for mrremind. Please don't do that.

R/readIIASA_subs_taxes.R
```diff
@@ -9,7 +9,6 @@
 #' pre final energy category \item \code{subsidies_bulk}: subsidy rate per final
 #' energy category \item \code{energy}: final energy quantities per category }
 #' @return magpie object of the IIASA_subs_taxes data
-#' @author Christoph Bertram
```
fbenke-pik, line 12: WHy did you remove the Author here? This is useful and important info, even if the person is no longer at PIK.

R/calcCO2Prices.R
```diff
@@ -14,17 +11,15 @@ calcCO2Prices <- function() {
...
-  # read data used for weight
+  # Read data used for weight
   ceds <- calcOutput("Emissions", datasource = "CEDS2024", aggregate = FALSE)
   ceds <- ceds[, , "Emi|CO2|Energy and Industrial Processes (Mt CO2/yr)"]
-  # for years in the future, use last year available from CEDS
+  # For years in the future, use last year available from CEDS
   ceds <- ceds[, pmin(getYears(x), max(getYears(ceds))), ]
   getYears(ceds) <- getYears(x)
 
-  return(list(
```
fbenke-pik, line 24: We agreed as a team to not use that rule and switched off that linter. Therefore, I don't think it makes sense to go through legacy code to change it.

R/calcCapTarget.R
```diff
@@ -1,104 +1,87 @@
 #' Capacity targets from two sources
+#'
 #' @description The capacity targets (GW)  at regional level are produced from two different databases-
-#' UNFCCC_NDC database, an update of the Rogelj 2017 paper (see readme in inputdata), and REN21 Global Renewables report
-#' The UNFCCC_NDC capacity targets are further broken down to conditional and unconditional targets.
-#' @author Aman Malik, Oliver Richters
```
fbenke-pik, line 5: Please bring back all the `@author` tags. They are useful to evaluate who to ask for help and if the function is rather new or legacy code.

R/readGini.R
```diff
@@ -9,59 +9,26 @@
 #' described in the article. The country codes are the World Bank codes.
 #'
 #' @return magpie object of the Gini data
-#' @author Bjoern Soergel
```
fbenke-pik, line 12: Please bring back author

fbenke-pik:
There are several files with only stylistic changes. In future PRs, please make a separate PR for these. When doing complex refactorings, as in `calcCapTarget`, `readExpertGuess`, `readUNFCCC_NDC`, run inputdata generation with those refactorings (and no expected changes in generated data) and ensure that all the inputdata is unchanged using `madrat::compareData`. I know you ran `compareData` for this PR as well, but the problem is that you mix actual data changes with refactorings that do not change data and therefore you cannot tell for sure if you did not break anything using `compareData`.

R/calcFEShares.R
fbenke-pik: Thank you for pointing this out, Michaja.  I just tested and noticed that madrat even does not recognize when you pass the same scenarios but in different order and will recalculate.

fbenke-pik:
> Side note: I see this warning popping up at in multiple calculations. It seems to come from something happening in calcOutput. How do I get rid of it?
> 
> ```
> WARNING:
> ~ Weight sum is 0, so cannot normalize and will return 0 for some aggregation targets. This changes the total sum of the magpie object! If this is really intended set zeroWeight = "allow", or "setNA" to return NA.
> ```

This is a known issue, occurs when running some aggregations on magclass objects where NAs in weights (or possibly other reasons) cause the sum to be not equal to its parts. Needs to be addressed per case and is a task in the RSE backlog.

R/readUNFCCC_NDC.R
```diff
@@ -1,107 +1,114 @@
...
+#' @details Country name is ISO coded. Capacity/Additional Capacity targets are in GW. Generation/Production targets
+#'  are in GWh.
 #' @return  magpie object
 #' @author  Aman Malik, Christoph Bertram, Oliver Richters, Sophie Fuchs, Rahel Mandaroux
 #' @param subtype Capacity_2023_cond (or 2018/2021/2022 or uncond) for capacity target,
-#' Emissions_2023_cond (or 2018/2021/2022 or uncond) for Emissions targets
-#' @importFrom readxl read_xlsx
-#' @importFrom dplyr select
+#'  Emissions_2023_cond (or 2018/2021/2022 or uncond) for Emissions targets
+#' @param subset A string (or vector of strings) designating the scenario(s) to be returned.
```
fbenke-pik, line 12: Maybe add a remark that it is needed for the convert function, as it is not used directly in the read function

R/convertEdgeBuildings.R
```diff
@@ -1,58 +1,12 @@
...
-    exceeding_years_after  <- exceeding_years_vec[exceeding_years_vec > threshold]
-    lambda <- c(rep(0, length(previous_years)),
-                utils::tail(seq(0, 1, length.out = length(exceeding_years_before) + 1), -1),
-                rep(1, length(exceeding_years_after)))
-    names(lambda) <- as.character(c(previous_years, exceeding_years_vec))
-    return(as.magpie(lambda))
-  }
-
-
+convertEdgeBuildings <- function(x, subtype, subset) {
```
fbenke-pik, line 9: `readEdgeBuildings` does not have the param 'subset`. Does it work then? If so, this is inconsistent with readUNFCCCC_NDR.R where you add the param to the read function (probably because you want to use it in the convert function)

johanneskoch94:
Thanks for looking everything over Falk! I'll reply to all your comments, but I'll probably only be able to set up all the approiate PRs in 2 weeks, when I'm back from vacation ;)

R/convertEdgeBuildings.R
johanneskoch94: Jup. Improves cache use, since calls with different subset arguments can use the same read cache. And yes, in readUNFCCC_NDR.R is incosistent - will fix.

R/readUNFCCC_NDC.R
johanneskoch94: OK!

R/readIIASA_subs_taxes.R
johanneskoch94: I apologize. I thought the consensus was that we should avoid using author tags and instead rely on git blame. Will be but back.

R/convertEdgeBuildings.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > Does it work then?

Yes.  The `read` and `convert` functions can have only the arguments `subtype` and `subset` (and `x`), but those are marshalled by `madrat` automatically.
I would argue to include the `subset` argument for the `read` function, too, as that will be the thing people are looking up.  (How many users do know the `convert` function is run automatically?)

fbenke-pik:
I am done with going through all of the changes. First of all, thanks for the improvement @johanneskoch94! It is a great step towards managing scenarios centrally. 

I hope that the concerns of @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q were solved, as he did not get the chance to approve before it is merged!? If not, please let us know if there are other open issues with this PR still.

Introducing the new argument to calc/read/convert functions seems necessary to manage scenarios centrally., but let's keep our eyes open for any instances where we accidentally violated the reuse of cached files.

R/readStationary.R
fbenke-pik: makes it hard to find functions for people who are browsing the repo online or are not familiar with commandline tools like `grep`

R/readIIASA_subs_taxes.R
fbenke-pik: We have users wanting to track sources, but do not know how to use git blame (because they are not coder themselves). Besides, people who added or changed code are not always the same as the person to contact with questions. Lavinia is listed in git blame for code that has not been changed since the migration to GitHub, so that does not help. Other legacy functions have been updated lately only by people how did formatting or refactorings.

johanneskoch94: Ok, but then surely it would be better to have actual contact people listed as authors, instead of people who don't work for PIK anymore no? (See #640 for the return of the authors tag.)

R/convertEdgeBuildings.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > Improves cache use, since calls with different subset arguments can use the same read cache.

Those requests hit the `convert`-cache directly.

But, this broke `mredgebuildings`: https://github.com/pik-piam/mredgebuildings/blob/0e61f5ddc659f34b973155a0eea2ea63c5e9074a/R/calcFloorspacePerCap.R#L14
FYI @robinhasse, @ricardarosemann

R/calcFEShares.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > I just tested and noticed that madrat even does not recognize when you pass the same scenarios but in different order and will recalculate

:point_right: https://github.com/pik-piam/mrindustry/pull/30#discussion_r1965141924

R/convertEdgeBuildings.R
johanneskoch94: Damn... I'm sorry! Will set up a PR.

R/calcFEShares.R
fbenke-pik: Ah, cool! From what I can tell, @johanneskoch94 has adapted this practice in this PR as well :+1:

R/convertEdgeBuildings.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: This concerns three function pairs currently (all introduced by this PR):

```r
library(madrat)

setConfig(.verbose = FALSE)

suppressPackageStartupMessages({
    installedMadratUniverse() |>
        sapply(library, character.only = TRUE) |>
        invisible()
})

for (convert in
     getCalculations(prefix = 'convert', globalenv = FALSE)[['call']]) {
    read <- sub(':::convert', ':::read', convert)

    no_read_function <- is.na(
        match(sub('.*:::', '', read),
              ls(envir = asNamespace(sub(':::.*', '', read)))))

    if (no_read_function) {
        next
    }

    conv_formals <- formals(eval(parse(text = convert))) |>
        Filter(f = is.symbol) |>   # w/o those that have default values
        names()
    read_formals <- formalArgs(eval(parse(text = read)))

    x <- setdiff(conv_formals, c('x', read_formals))

    if (length(x)) {
        warning(cli::pluralize(
            read, ' misses argument{?s} {paste0("`", x, "`")} ',
            'that have no defaults in ', convert), call. = FALSE)
    }
}

# Warning messages:
# 1: mrremind:::readEdgeBuildings misses argument `subset` that have no defaults in mrremind:::convertEdgeBuildings 
# 2: mrremind:::readNewClimate misses argument `subset` that have no defaults in mrremind:::convertNewClimate 
# 3: mrremind:::readStationary misses argument `subset` that have no defaults in mrremind:::convertStationary 
```
But only `edgeBuildings` is negatively affected
```
$ reposearch 'readSource.*(EdgeBuildings|NewClimate|Stationary)'
repos/pik-piam/mrremind/master/R/calcFloorspace.R
18:  data <- readSource("EdgeBuildings", subtype = "Floorspace", subset = scenario)

repos/pik-piam/mrremind/master/R/calcFeDemandBuildings.R
17:  stationary <- readSource("Stationary", subset = scenario)
18:  buildings  <- readSource("EdgeBuildings", subtype = "FE", subset = scenario)

repos/pik-piam/mrremind/master/R/calcCapTarget.R
24:    capCond <- readSource("NewClimate", subtype = "Capacity_2025_cond")
25:    capUncond <- readSource("NewClimate", subtype = "Capacity_2025_uncond")

repos/pik-piam/mrremind/master/R/calcEmiTarget.R
63:      "2025_cond"   = readSource("NewClimate", subtype = "Emissions_2025_cond", subset = subsetScen),
64:      "2025_uncond" = readSource("NewClimate", subtype = "Emissions_2025_uncond", subset = subsetScen)

repos/pik-piam/mredgebuildings/main/R/calcFloorspacePerCap.R
14:  fs <- readSource("EdgeBuildings", "Floorspace") %>%

repos/pik-piam/mrindustry/main/R/calcFeDemandIndustry.R
51:  stationary <- readSource("Stationary", subset = remind_scenarios)[, , c("feindheat", "feh2i")]
```
as the other two packages where handled (and tested during input data generation).

Funny this apparently never before happened in ten years or so …

Opinions:
- > It makes debugging and reading code sooo much easier to have the read and convert functions in the same file

  Agreed.
- > makes it hard to find functions for people who are browsing the repo online or are not familiar with commandline tools like `grep`

  I do not see that.  Both RStudio and VSCode have <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd> for "Find in Files".  If one cannot manage that, they probably have no business messing with the source in the first place.  Same for the search function on Github:
  image
  This could be a clear and simple rule: `convert` functions are in the file of the corresponding `read` function.

- I think what might be much more of a stumbling block is the level of documentation concerning `readSource()` and the existence of `convert` functions.  I think most people puzzle that out for themselves instead of gleaning it from the `madrat` vignette -- I sure did.
- I also think this would be improved when `read` and `convert` (and probably `correct`) functions were documented together and when `convert` functions, as a rule, would inherit the parameters of the `read` function.
  ```diff
   #' Convert EDGE Buildings data to data on ISO country level.
   #'
  -#' @param subtype either FE or Floorspace
  -#' @param subset A string (or vector of strings) designating the scenario(s) to be returned.
  +#' @rdname readEdgeBuildings
   #' @param x MAgPIE object containing EDGE values at ISO country resolution
  +#' @inheritParams readEdgeBuildings
   #' @return EDGE data as MAgPIE object aggregated to country level
   #' @author Antoine Levesque, Robin Hasse
   ```
  image
  This way, in order to have the parameters documented in the `convert` function, they must be documented and thus declared in the `read` function, too.
  But that would be a group activity of course, and is beyond this issue.

R/calcFEdemand.R
```diff
@@ -1,17 +1,16 @@
...
+calcFEdemand <- function(scenario) {
 
-  feBuildings <- calcOutput("FeDemandBuildings", subtype = "FE", warnNA = FALSE, aggregate = FALSE)
-  feIndustry <- calcOutput("FeDemandIndustry", warnNA = FALSE, aggregate = FALSE)
+  feBuildings <- calcOutput("FeDemandBuildings",
+                            subtype = "FE",
+                            scenario = scenario,
+                            warnNA = FALSE,
+                            aggregate = FALSE)
+  feIndustry <- calcOutput("FeDemandIndustry",  scenario = scenario, warnNA = FALSE, aggregate = FALSE)
```
fbenke-pik, line 11: @LaviniaBaumstark @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q @johanneskoch94 This should be `scenarios` as an argument, not `scenario`, right? This should explain the problems with caching we are experiencing currently.

fbenke-pik: https://github.com/pik-piam/mrremind/pull/642

fbenke-pik: @johanneskoch94 I was in a rush to fix this. Maybe you want to solve this differently by rather renaming the argument in `calcFeDemandIndustry` to `scenario` for improved consistency.

fbenke-pik: Fine by me, just be aware that it is `scenario` in mrremind - and madrat won't necessarily tell you if you get it wrong :)

PR #635, opened 2025-02-21
Add near-term electricity demand to buildings in CHA

robinhasse:
Electricity demand in CHA in 2025 is to low. We want to tackle this in EDGE-B but need a hot fix now. I increase the non-specified electricity demand that we add to the buildings demand by 2EJ/yr in the near future.

Looking at it I realised again that we should really look into what we do with stationary. The non-specified electricity demand in CHA has continuously increased over the last year reaching 1.75 EJ/yr in 2020. The very old numbers in stationary are only based on observations until 2010 and assume a decline from there. In `convertStationary`, we overwrite history with recent IEA data and the linearly transition back from the last IEA period (this is where I added 2 EJ/yr) to the projection until 2035. Where I added demand, I shifted the transition to start from 2025 instead of 2020 and extend until 2055 to make the peak less pronounced.

But it is clear, that our very ld projection does not continue observed historic trends.

Hopefully, we can add demand in EDGE-B and remove this hotfix in about 2 weeks from now.

grafik
*please ignore the historic deviation of some (partly removed) scenarios. This likely results from using an outdated cache.*

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '43903020'
+ValidationKey: '43925340'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


ricardarosemann (approved):
Looks ok, thanks!

PR #647, opened 2025-03-12
remove unused subtype = "ch4_history" to read/convertCEDS2024

pascal-sauer:


gabriel-abrahao:
I'm a bit puzzled, that line was originally mine, and looking at it now I can't understand how it worked in the first place. Does `readSource` ignore a `subtype` parameter if it's unused?

Either way, thanks for the catch!

gabriel-abrahao (approved):

PR #653, opened 2025-03-21
Increase short-term ONONSPEC electricity projection in CHA

robinhasse:
To increase electricity demand from buildings in CHA in 2025, this PR increases the respective ONONSPEC projection. In this course, I also changed the regression for the end-of-data coefficients from a linear to a spline regression. This should help to capture non-linear trends. The impact on results is very small though:

ononspecComparison_mainpulateCHN

The most notable change is the specifically manipulated electricity demand in CHA that should now close the gap to the 2025 target demand derived from extrapolating recent growth.

grafik

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '44602768'
+ValidationKey: '44631784'
 AcceptedWarnings:
 - Invalid URL: .*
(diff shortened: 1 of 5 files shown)


ricardarosemann (approved):
Thank you, looks good!

PR #657, opened 2025-03-24
move calc functions to fullVALIDATIONREMIND

fbenke-pik:
This moves all the remaining calc functions from calcHistorical to fullVALIDATIONREMIND and creates smaller calc functions where suitable.


R/calcEmissions.R
```diff
@@ -1127,8 +1127,8 @@ calcEmissions <- function(datasource = "CEDS16") {
     }
 
     tmp <- tmp * 44 / 12 / 1000 # from ktC -> MtCO2
-
```
fbenke-pik, line 1130: This is the only time when I dropped data before 1970 from historical.mif, because the description below says so

R/calcEEAGHGEmissions.R
pweigmann, line 1: Technically, I think madrat would prefer this in ``calcEmissions.R`` instead of having calc functions for individual data sources, at least when the function is short. Unless there is a special reason to do it this way, maybe somehow because of regional resolution?

pweigmann (approved):
Looks good and a very welcome effort, thanks! Didn't have time to test mif generation and data integrity though.

LaviniaBaumstark (approved):

PR #659, opened 2025-03-25
Update capacity factors in China

robertpietzcker:
To produce a reasonable coal power deployment in CHA from 2020 to 2025 (increase), some adjustments to CFs are needed. 

a) Hydro production is lower
b) coal CFs don't go down as quickly as expected
c) PV generation is a bit lower


LaviniaBaumstark (approved):

.gitignore
```diff
@@ -8,3 +8,4 @@ figures
 /TODO.txt
 output
 mrremind.Rproj
+mrremind_rob.Rproj
```
fbenke-pik, line 11: Instead of adding your own file name, just change the line above to `*.Rproj`

fbenke-pik (approved):

.gitignore
robertpietzcker: thanks, good to know!

PR #665, opened 2025-03-28
nmove ldv transport shares to separate calc functio

fbenke-pik:
to avoid unnecessary calculations


fbenke-pik:
I checked tht `f04_IO_input.cs4r` and `f04_IO_output.cs4r` do not change by this refactoring.

johannah-pik (approved):
Looks good! Thanks for fixing that

PR #669, opened 2025-04-07
Renewable capacity lower-bounds until 2025 

lecfab:
- See remind side of this PR: https://github.com/remindmodel/remind/pull/2080
- After Pascal did the https://github.com/pik-piam/mrremind/pull/666, we have renewable capacity until 2023.

### Historical and near-term trends
- Historical capacity is now smoothed over 5-year periods (until 2022), to match Remind behaviour.
- 2023 IRENA renewable capacities gives a lower-bound for 2025.
- I tried to use GlobalEnergyMonitor to obtain projected values to 2025, but it was not convincing. We could alternatively enter manually some of the 2025 values.

### Specific technologies
- wind and solar details are now included in historical.mif using IRENA values
- hydro historical capacity is now given at technology level instead of primary energy level (pm_histCap instead of p_PE_histCap).
- hydro maxprod had to be adjusted so that NDC targets stay feasible in EU21 runs: UKI has factor 2.1 instead of 1.625, and AUT (Austria) has factor 1.3 like its fellow EWN countries instead of 1.
- geohdr potential is now based on 2020 capacity instead of 2015. A multiplier x3 was transferred from remind to mrremind


pweigmann (approved):
Looks good to me in general, can't say much about what it means for REMIND.

PR #670, opened 2025-04-08
fix emission variables wrt bunkers

pweigmann:
After the default behavior of REMIND variables was changed to include bunkers if not specified otherwise, this PR harmonizes CEDS2024, UNFCCC and EDGARghg emission data accordingly. Other emissions data sources probably need to be adjusted as well if we want to continue using them. @fbenke-pik This might be true for Primap and Eurostat data .

Accompanying remind2 PR to adjust the variable names in emission plots: https://github.com/pik-piam/remind2/pull/724

A test of the ``historical.mif`` generation can be found as ``rev7.37bunkerEmi`` on the cluster.

@johanneskoch94 I adjusted ``calcCO2prices`` to use the new name so that the values should stay the same

@gabriel-abrahao I am not sure what this means for [``calcEmissions4ReportExtra.R``](https://github.com/pik-piam/mrremind/blob/96370aeb5b192efa345aa0f69593f01862154921/R/calcEmissions4ReportExtra.R), so I didn't do any changes there. Let me know if you don't know what to do, be aware that also the mapping was changed.

For both sources the following variables are now available (CO2 example, same for other pollutants):

**Bunker variables**
`Emi|CO2|Energy|Demand|Transport|International Bunkers`
`Emi|CO2|Transport|Freight|International Shipping|Demand`
`Emi|CO2|Transport|Pass|Aviation|International|Demand`

`Emi|CO2|w/ Bunkers|Energy`
`Emi|CO2|w/ Bunkers|Energy and Industrial Processes`
`Emi|CO2|w/ Bunkers|Energy|Demand`
`Emi|CO2|w/ Bunkers|Energy|Demand|Transport`

**Changed emission variables**

Used to be without bunkers, now with bunkers!
`Emi|CO2|Energy`
`Emi|CO2|Energy and Industrial Processes`
`Emi|CO2|Energy|Demand`
`Emi|CO2|Energy|Demand|Transport`

Didn't exist before
`Emi|CO2|w/o Bunkers|Energy`
`Emi|CO2|w/o Bunkers|Energy and Industrial Processes`
`Emi|CO2|w/o Bunkers|Energy|Demand`
`Emi|CO2|w/o Bunkers|Energy|Demand|Transport`

@RahelMA and others who might need this. The latest [commit](https://github.com/pik-piam/mrremind/pull/670/commits/b581032f81c0b9a42275ad621691baa7b8df6e0d) adds ``LULUCF national accounting`` to the land-use (containing) variables. Hope I understood correctly how to do this and this is now consistent with [REMIND variables](https://github.com/pik-piam/remind2/blob/c2897d866c901294cf364770ae86dc8675c7e351/R/reportEmi.R#L2933).


inst/extdata/sectoral/mappingCEDS2024toREMIND.csv
```diff
@@ -17,13 +17,13 @@ CEDS2024;REMIND;IAMC;CommentREMIND;CommentIAMC
 1A2g_Ind-Comb-transpequip;Energy|Demand|Industry|Other;Industrial Sector;;
 1A2g_Ind-Comb-wood-products;Energy|Demand|Industry|Other;Industrial Sector;;
 1A3ai_International-aviation;Transport|Pass|Aviation|International|Demand;Aircraft;Transport emissions variables from EDGE-T reporting do not comply with emissions reporting structure so far, once REMIND names adapted -> adapt here as well;
-1A3aii_Domestic-aviation;Energy|Demand|Transport;Aircraft;;
-1A3b_Road;Energy|Demand|Transport;Transportation Sector;;
-1A3c_Rail;Energy|Demand|Transport;Transportation Sector;;
+1A3aii_Domestic-aviation;w/o Bunkers|Energy|Demand|Transport;Aircraft;;
+1A3b_Road;w/o Bunkers|Energy|Demand|Transport;Transportation Sector;;
```
fbenke-pik, line 21: is this really the way the variable should be called?

R/fullVALIDATIONREMIND.R
```diff
@@ -136,11 +136,28 @@ fullVALIDATIONREMIND <- function(rev = 0) {
...
-  )
+  ceds <- calcOutput(
+    "Emissions", datasource = "CEDS2024",
+    aggregate = columnsForAggregation, warnNA = FALSE, try = FALSE)
+
+  # the following variables only have meaningful data on global level
+  # as bunkers exist in source only on the global level and were only added to
+  # regions because madrat doesn't allow global data before this point
+  # (see mrcommons::convertCEDS2024)
+  vars_glo_only <- getNames(ceds)[grepl("w/ Bunkers", getNames(ceds))]
```
fbenke-pik, line 147: if possible, keep fullVALIDATION as short as possible. no extra calculations here, unless absolutely necessary. in an ideal world, this is just a list of called calc functions with settings.

R/fullVALIDATIONREMIND.R
```diff
@@ -136,11 +136,28 @@ fullVALIDATIONREMIND <- function(rev = 0) {
...
-    try = FALSE, writeArgs = list(scenario = "historical", model = "CEDS")
-  )
+  ceds <- calcOutput(
+    "Emissions", datasource = "CEDS2024",
+    aggregate = columnsForAggregation, warnNA = FALSE, try = FALSE)
+
+  # the following variables only have meaningful data on global level
+  # as bunkers exist in source only on the global level and were only added to
+  # regions because madrat doesn't allow global data before this point
+  # (see mrcommons::convertCEDS2024)
```
fbenke-pik, line 146: we have encountered similar problems before, and there is a solution that can be applied to the calc function directly. see `calcIEA_ETP`

inst/extdata/sectoral/mappingCEDS2024toREMIND.csv
pweigmann: The mapping leaves out the ``Emi|<pollutant>`` prefixes, so the final names after the calc function will be the same as on REMIND side after this PR: https://github.com/pik-piam/remind2/pull/709

gabriel-abrahao (approved):
Thanks for flagging this to me! But since it appears to only affect CO2 emissions, `reportExtraEmissions` should be fine, it only deals with CH4 and N2O.

pweigmann:
> Thanks for flagging this to me! But since it appears to only affect CO2 emissions, `reportExtraEmissions` should be fine, it only deals with CH4 and N2O.

Ah, sorry no, it affects all pollutants! Just used CO2 as example.

gabriel-abrahao:
> > Thanks for flagging this to me! But since it appears to only affect CO2 emissions, `reportExtraEmissions` should be fine, it only deals with CH4 and N2O.
> 
> Ah, sorry no, it affects all pollutants! Just used CO2 as example.

Oh, I see. But it's fine too, that script currently has its own `calc` function for its inputs, `calcEmissions4ReportExtra`, which calls `readCEDS2024` directly.

pweigmann:
> > > Thanks for flagging this to me! But since it appears to only affect CO2 emissions, `reportExtraEmissions` should be fine, it only deals with CH4 and N2O.
> > 
> > 
> > Ah, sorry no, it affects all pollutants! Just used CO2 as example.
> 
> Oh, I see. But it's fine too, that script currently has its own `calc` function for its inputs, `calcEmissions4ReportExtra`, which calls `readCEDS2024` directly.

That's good, just be mindful, that also the mapping is being changed.

PR #673, opened 2025-04-14
Remove Warnings and fix bug in fullREMIND

fbenke-pik:


R/calcCCScapacity.R
```diff
@@ -9,7 +9,9 @@
 #'
 #' @export
 calcCCScapacity <- function(subtype) {
-  x <- calcOutput("ProjectPipelines", subtype = "CCS", aggregate = F)
+
+  x <- calcOutput("ProjectPipelines", subtype = "CCS",
+                  aggregate = FALSE, warnNA = FALSE)
```
LaviniaBaumstark, line 14: are we sure, that NAs are o.k. for this function?

LaviniaBaumstark (approved):

R/fullREMIND.R
```diff
@@ -103,14 +103,14 @@ fullREMIND <- function() {
...
+  calcOutput("EmiFossilFuelExtr", source = "CEDS2025",                          round = 6, file = "p_emiFossilFuelExtr2020.cs4r")
   calcOutput("Region2MAGICC",                                                   round = 6, file = "p_regi_2_MAGICC_regions.cs3r")
   calcOutput("EmiPollutantExo", subtype = "Waste",                              round = 6, file = "f11_emiAPexo.cs4r")
   calcOutput("EmiAirPollLandUse",                                               round = 6, file = "f11_emiAPexoAgricult.cs4r")
   calcOutput("GAINSEmi", subtype = "emissions",                                 round = 5, file = "emi_gains.cs4r")
   calcOutput("GAINSEmi", subtype = "emission_factors",                          round = 5, file = "ef_gains.cs4r")
   calcOutput("GAINSEmi", subtype = "emissions_starting_values",                 round = 5, file = "f11_emiAPexsolve.cs4r")
-  calcOutput("EmissionFactors", subtype = "emission_factors", warnNA = TRUE,    round = 5, file = "f11_emiFacAP.cs4r")
+  calcOutput("EmissionFactors", subtype = "emission_factors", warnNA = FALSE,   round = 5, file = "f11_emiFacAP.cs4r")
```
fbenke-pik, line 113: We should probably validate as well, if NAs are ok in this one. But not sure who is the "owner" of this function. @LaviniaBaumstark  any idea who to ask?

LaviniaBaumstark: hmm, maybe , @fschreyer can answer this question?

fschreyer: I have never seen this function. Seems like these are emissions factors for air pollutants used in module 11. @dbantje would need to say whether that's problematic.

R/calcCCScapacity.R
pweigmann: Yes, from my pov thats ok.

R/fullREMIND.R
dbantje: I might be able to say with some digging, but I think @gabriel-abrahao knows more about module 11 than me actually.

PR #675, opened 2025-04-29
Remove weight sum warnings

fbenke-pik:
This PR primarily deals with the warning 

`Weight sum is 0, so cannot normalize and will return 0 for some aggregation targets. This changes the total sum of the magpie object! If this is really intended set zeroWeight = "allow", or "setNA" to return NA.`

This PR deals with these warning in two ways

A) Set the weight to very small values

This was done whenever setting small values would not change the generated input data, as the actual values for the affected regions are zero anyways. 

B) Allow NAs in weights

This was done whenever setting small values actually would change the generated input data for some regions. In most of these cases, the principal authors of these functions are no longer available and cannot be asked if this is the best way to proceed. So we assume that zero weights are ok, as the produced input data is what they intended.


R/calcTransportGDPshare.R
```diff
@@ -31,8 +31,8 @@ calcTransportGDPshare <- function() {
   mapGLO$RegionCode <- "GLO"
   w <- ggdc10[, , "VA_Q05.Summation of sector GDP"]
   share <- share %>%
-    toolFillWithRegionAvg(valueToReplace = 0, weight = w, regionmapping = mapH12) %>%
```
fbenke-pik, line 34: @Renato-Rodrigues 

This silences the following warning. If you think we should rather address it, let me know.
```
~ WARNING: 
~ More than 50% missing values for: LAM, OAS, SSA, EUR, NEU, MEA, REF, CAZ
```

R/fullREMIND.R
```diff
@@ -133,12 +133,8 @@ fullREMIND <- function() {
   calcOutput("IO",   subtype = "input",               round = 8,  file = "f04_IO_input.cs4r")
   calcOutput("IO",   subtype = "trade",               round = 8,  file = "f_IO_trade.cs4r")
   calcOutput("Clinker_to_cement_ratio",               round = 2,  file = "p37_clinker-to-cement-ratio.cs3r")
-  # delete the 'dummy' line
-  system(paste0('sed -i "/dummy/d" ', getConfig()$outputfolder, "/p37_clinker-to-cement-ratio.cs3r"))
 
-  avg5years <- quitte::remind_timesteps # average over 5y before 2023
-  avg5years <- avg5years[(avg5years$period <= 2020) | avg5years$period == avg5years$year,,]
```
fbenke-pik, line 140: @lecfab madrat takes care of adjusting the temporal mapping as you intend here by itself.

R/readGEA2012.R
```diff
@@ -177,7 +177,7 @@ readGEA2012 <- function(subtype) {
     # Disaggregate the GEA data according to the BGR data on country-level oil/gas combined reserves + resources
     w <- dimSums(w, dim = 3)
     sp_IEADecRat <- toolAggregate(sp_IEADecRat, mappingREM11, weight = NULL)
-    sp_IEADecRat <- toolAggregate(sp_IEADecRat, mappingGEA, weight = w)
```
fbenke-pik, line 180: I allowed the zero weight to not change the inputdata, as the author Stephen Bi is no longer around.

R/calcGAINSEmi.R
```diff
@@ -64,27 +63,30 @@ calcGAINSEmi <- function(subtype = "emissions") {
...
     unit <- "Mt/TWa"
     description <- "Emission factors from GAINS for all CEDS setors execpt aviation and shipping"
-    weight <- w
-  }
 
-  if (subtype == "emissions_starting_values") {
+    weight <- emi_gains
+
+    # do not throw a warning for zero weights, as they seem to be intended
+    aggregationArguments <- list(zeroWeight = "allow")
```
fbenke-pik, line 74: Allow zero weight to suppress warning and not change input data (author Sebastian Rauner is no longer with us)

R/calcFETaxes.R
```diff
@@ -120,12 +120,16 @@ calcFETaxes <- function(subtype = "taxes") {
...
   # the tax in a Cooling country is very high and the tax in a country in the
   # same region using a lot of electricity for cooking is low, the tax for
   # cooling and cooking with electricity will be equal where it should be high
   # for cooling and low for cooking
   # So, we can assume that countries are app. similar in a given region
 
-  list(x = Rtax, weight = Renergy, unit = "US$2017/GJ", description = desc)
+  list(x = Rtax, weight = Renergy, unit = "US$2017/GJ", description = desc,
+       # do not throw a warning for zero weights, as they seem to be intended
+       aggregationArguments = list(zeroWeight = "allow")
```
fbenke-pik, line 133: Allow zero weights to suppress warnings without changing input data. @Renato-Rodrigues LEt me know if this should be dealt with differently (e.g. by using small weights != 0)

fbenke-pik: Talked to Renato and he approved of the current approach.

R/calcTransportGDPshare.R
fbenke-pik: This file can be removed altogether according to Renato

R/fullREMIND.R
lecfab: really? at the beginning i simply put `avg5years <- quitte::remind_timesteps`, but it caused issues.
For instance, if a technology has historical value for 2023 and a manual value for 2025, then the 2023-2027 averaging gives unexpected results.

But maybe it works with the updated default behaviour!
Would you mind checking that the result is exactly the same with and without these lines?

fbenke-pik: I checked and inputdata did not change for `pm_histCap.cs3r` after I adjusted the code.

The case you describe never occurs with the current data, because `calcCapacity` with subtype `capacityByTech` only creates the years 

```[1] "y2000" "y2001" "y2002" "y2003" "y2004" "y2005" "y2006" "y2007" "y2008" "y2009" "y2010" "y2011" "y2012" "y2013" "y2014"
[16] "y2015" "y2016" "y2017" "y2018" "y2019" "y2020" "y2021" "y2025"```

lecfab: Thanks for checking!

lecfab (approved):

PR #680, opened 2025-05-19
Base for updating air pollutant emissions in REMIND

gabriel-abrahao:
Consider this a draft for now. Adds functions to calculate emission factors for air pollutants based on the latest data gotten from Zig and Shaohui at IIASA, here called `GAINS2025`. The implementation attempts to keep all outputs compatible with the previous implementation, here called `GAINSlegacy`, with the addition of an `ssp` dimension.

GAINS2025 has projections for different SSPs by design for all scenarios, and that's also taken into account for the 2050-2100 extension and their weights. 

## Tasks
- [x] Emissions and emission factors at the GAINS sector level
- [x] Starting values for emissions in REMIND (decided to keep old values for now)
- [x] Emission factors at the REMIND sector/technology level
- [x] OPTIONAL: Output CEDS2020 emissions for hamonization
- [x] Exogenous Waste emissions (decided to keep old values for now)
- [x] Exogenous landuse emissions (decided to keep old values for now)
- [x] Exogenous Aviation and Shipping emissions (decided to keep old values for now)
- [x] `fullREMIND`
- [x] Smooth transition in scenarios
- [x] Complete testing in REMIND

## Plots for review
Please take a look @nicobauer , if possible also those who (a very long time ago) seem to have worked with this @strefler @dklein-pik 

**Emissions and emission factors at the GAINS sector level**
For all available GAINS2025 scenarios:
[rawefs_agg.pdf](https://github.com/user-attachments/files/20306494/rawefs_agg.pdf)
[rawemi_agg.pdf](https://github.com/user-attachments/files/20306478/rawemi_agg.pdf)

EF comparisons with GAINSlegacy for SSP2 and two scenarios in common (CLE and MFR):
[comp_agg.pdf](https://github.com/user-attachments/files/20306513/comp_agg.pdf)

**Emission factors at the REMIND sector/technology level**
[remefs_all.pdf](https://github.com/user-attachments/files/20306521/remefs_all.pdf)

**Global total emissions comparison between GAINS versions and CEDS (in 2020)**
[comptot.pdf](https://github.com/user-attachments/files/20306528/comptot.pdf)


gabriel-abrahao:
@laurinks thanks for taking a look! Keep in mind that the emissions in those plots are *not* what will eventually come out of REMIND, but just the GAINS scenario emissions. Those are used just for some exogenous sectors, and of course to derive the emission factors, but are a good indicator of the general change to be expected in REMIND. The actual REMIND emissions will depend on REMIND's activities, but my goal is to rescale the emission factors to get the output close to CEDS afterwards.

laurinks:
Excellent reply, thank you @gabriel-abrahao ! It is a pitty that the WG on Air Pollution in ScenarioMIP was so inactive and not actively managed by IIASA. If there would have been regular meetings, I believe that many of the questions would have been resolved by now. Anyway, no point in complaining ;)

Thanks for the concrete and pragmatic suggestions. Somewhat difficult for me to make a qualified judgement without discussing the pros and cons. I would suggest an offline discussion tomorrow, @nicobauer and @gabriel-abrahao.

gabriel-abrahao:
Here are the same plots as in the description comment, but with the smoothing applied and the new VLE scenario (see `calcGAINS2025scenarios.R` for a description).

[rawefs_agg.pdf](https://github.com/user-attachments/files/20490803/rawefs_agg.pdf)
[remefs_all.pdf](https://github.com/user-attachments/files/20490797/remefs_all.pdf)
[compemi_agg.pdf](https://github.com/user-attachments/files/20490799/compemi_agg.pdf)
[compefs_agg.pdf](https://github.com/user-attachments/files/20490800/compefs_agg.pdf)
[rawemi_agg.pdf](https://github.com/user-attachments/files/20490802/rawemi_agg.pdf)

laurinks (approved):

R/calcAirPollEmiRef.R
```diff
@@ -0,0 +1,94 @@
...
+#'
+#'
+#' @param baseyear year to take as a reference from CEDS, ignored for the EDGAR2005 LUC CO2 emissions
+#' @return magclass object
+#' @author Gabriel Abrahao
+#' @importFrom magclass getNames<- getYears<-
+
+calcAirPollEmiRef <- function(
+    subtype = "total", baseyear = 2020, outunits = "Mt/yr", namesformat = "GAINS2025") {
+  # require(tidyverse)
```
fbenke-pik, line 12: Please always remove commented out code.

R/calcGAINS2025forREMIND.R
```diff
@@ -0,0 +1,177 @@
...
+
+    # GAINSlegacy weights were for a fixed year, expand the time dimension
+    # so that it can be merged with the new data
+    wgtold <- mbind(lapply(getYears(wgtnew), \(x) setYears(wgtold, x)))
+
+    # ==============================================================================================================
+    # Combining GAINSlegacy and GAINS2025
+    # ==============================================================================================================
+    out <- mbind(outold, outnew)
+    wgt <- mbind(wgtold, wgtnew)
```
fbenke-pik, line 154: The chosen weight produces the warning `Weight sum is 0, so cannot normalize and will return 0 for some aggregation targets. This changes the total sum of the magpie object! If this is really intended set zeroWeight = "allow", or "setNA" to return NA.`

Basically, what happens is that when summing up the weights for some regions, the weigtht sum is zero and therefore, values that were not zero can get lost in the process of regional aggregation. Is this ok for you?

R/calcGAINS2025forREMIND.R
```diff
@@ -0,0 +1,177 @@
...
+    unit <- getFromComment(innew, "unit")
+  } else if (subtype == "emission_factors") {
+    linold <- calcOutput("GAINSEmi", subtype = "emission_factors", aggregate = F, supplementary = T)
+    inold <- linold$x
+    linnew <- calcOutput("GAINS2025scenarios", subtype = "emission_factors", aggregate = F, supplementary = T)
+    innew <- linnew$x
+    out <- bindNewOld(innew, inold)
+
+    # Weights, GAINS2025 uses activities that are per SSP but not per scenario, pad the dimensions
+    wgt <- bindNewOld(linnew$w, linold$w)
```
fbenke-pik, line 63: The chosen weight causes two warnings

```
1: In toolAggregate(x = x$x, weight = x$weight, rel = map$rel) :
  Negative numbers in weight. Dangerous, was it really intended?
2: In toolAggregate(x = x$x, weight = x$weight, rel = map$rel) :
  Weight sum is 0, so cannot normalize and will return 0 for some aggregation targets. This changes the total sum of the magpie object! If this is really intended set zeroWeight = "allow", or "setNA" to return NA.
  ```

Is this really intended, or are these actual problems?

**Comment by robertpietzcker** [comment]
@gabriel-abrahao @laurinks 
I just took a look at the data ([remefs_all.pdf](https://github.com/user-attachments/files/20490797/remefs_all.pdf)) and there are a number of things I don't understand/ find implausible:

assuming that "SSP2" is the new data: 
- why does CHA have much lower SO2 emissons from pecoal.sesofos.coaltr.indst than EUR? :
image

? I can hardly believe that current environmental legislation on coal burning is tighter in CHA than in the EU. 

(also, why is the MFR line higher than the CLE line?)

Similarly, for  pecoal.seel.pc:

image

why is CHA better than EUR from 2020 onwards?

File: R/calcGAINS2025forREMIND.R [hunk 4]
```diff
@@ -0,0 +1,177 @@
...
+#' files needed for REMIND in the proper format, for the scenario logic,
+#' see `GAINS2025scenarios`.
+#'
+#' @return Activity levels, emissions or emission factors
+#' @author Gabriel Abrahao
+#' @param subtype "emission_factors", "emissions","emissions_starting_values"
+#'
+#' @importFrom abind abind
+#' @importFrom utils head tail
+#' @importFrom magclass as.magpie
```
fbenke-pik, line 13: No need to import magclass functions.

R/calcAirPollEmiRef.R
```diff
@@ -0,0 +1,94 @@
+#' Calculate air pollutant emissions for a reference year, for use
+#' in combination with GAINS data at different sectoral aggregations
+#'
+#'
+#' @param baseyear year to take as a reference from CEDS, ignored for the EDGAR2005 LUC CO2 emissions
+#' @return magclass object
+#' @author Gabriel Abrahao
+#' @importFrom magclass getNames<- getYears<-
```
fbenke-pik, line 8: No need to import magclass functions

R/calcGAINS2025forREMIND.R
```diff
@@ -0,0 +1,177 @@
...
+#' @importFrom tidyr pivot_longer drop_na
```
fbenke-pik, line 14: Some of these are not used in the functions after all and seem to be leftovers from development

DESCRIPTION
```diff
@@ -40,6 +41,7 @@ URL: https://github.com/pik-piam/mrremind
 Depends:
     R (>= 4.0)
 Imports:
+    abind,
```
fbenke-pik, line 44: Are you sure you need to import a new library for what you want to achieve. Most likely there is a madrat-way

R/calcAirPollEmiRef.R
```diff
@@ -0,0 +1,94 @@
...
+
+calcAirPollEmiRef <- function(
```
fbenke-pik, line 10: Please add documentation for parameters `subtype`, `outunits` and `namesformat`

fbenke-pik: Consider dropping `outunits`, as it  always is "Mt/yr"

R/readGAINS2025.R
```diff
@@ -0,0 +1,79 @@
...
+#' @return Activity levels, emissions or emission factors. Alternatively,
+#' a Govenrnmment Capacity Index (GCI) used for deriving some scenario
+#' extensions.
+#' @author Gabriel Abrahao
+#' @param subtype "emission_factors", "emissions","emissions_starting_values"
+#'
+#' @importFrom magclass as.magpie
+#' @importFrom tidyr pivot_longer drop_na
+#' @importFrom readxl read_excel
+readGAINS2025 <- function(subtype, subset = "baseline.det") {
```
fbenke-pik, line 17: Please add documentation of the `subset` parameter

gabriel-abrahao:
Thanks for checking @robertpietzcker ! On your points:

> * why does CHA have much lower SO2 emissons from pecoal.sesofos.coaltr.indst than EUR? :
I don't have the expertise to judge here, but that's coming straight from the original GAINS data. See "Power_Gen_Coal" below. EFs for WEST_EURO are lower-ish than CHIN_PLUS, but it looks like IIASA doesn't have a very rosy view of CENT_EURO's coal generation. Note that their definition of "Central Europe" is what I'd call "Eastern Europe", i.e. everything east of Germany (which is West) until UKRA_PLUS (see the [mapping file](https://github.com/pik-piam/mrremind/blob/c80e56ed04ae28d24c2ef46b327a9ed31740c826/inst/extdata/regional/regionmapping_GAINS2025.csv) and [IMAGE's map](https://models.pbl.nl/image/Region_classification_map)).

Maybe there are some soviet-era coal plants that emit a lot of SO2 in former East countries that are part of our EUR? No idea. @nicobauer maybe it's worth raising it with Zig?

gains_comp

> (also, why is the MFR line higher than the CLE line?)

That's indeed odd, and seems to be happening somewhere in my calculations. I'll check the many good points raised by @fbenke-pik and see if it goes away.

fbenke-pik:
>  I'll check the many good points raised by @fbenke-pik and see if it goes away.

95% of the things I raised are just code conventions, nothing that should change data. Only the warnings could possibly affect data outcome.

R/calcGAINS2025forREMIND.R
```diff
@@ -0,0 +1,177 @@
...
+      subtype, " is not yet implemented, use instead the old starting values implementation from calcGAINSemi"
+    ))
+  } else if (subtype == "emission_factors_remindsectors") {
+    # ==============================================================================================================
+    # GAINS2025
+    # ==============================================================================================================
+    # Input emission factors
+    linnew <- calcOutput("GAINS2025scenarios", subtype = "emission_factors", aggregate = F, supplementary = T)
+    innew <- linnew$x
+    wgtnew <- linnew$w
```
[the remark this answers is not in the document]
gabriel-abrahao: That's odd, I don't get any warning on my setup, do you have a special setting to increase R verbosity? Either way, thanks for the heads up, I never even realized that it's not just called `w`, fixing.

R/calcAirPollEmiRef.R
gabriel-abrahao: Documented. `kt/yr` is not used now but I plan to, for harmonizing some emissions to CEDS later

R/calcAirPollEmiRef.R
```diff
@@ -0,0 +1,94 @@
...
+    mag <- mag[, , polnamesmap]
+    getItems(mag, "species") <- names(polnamesmap)
+    return(mag)
+  }
+
+  if (namesformat %in% c("REMIND","REMINDexo")) {
+    fullceds <- fixPolNames(fullceds, namesformat)
+  }
+
+  if (subtype == "total") {
```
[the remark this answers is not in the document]
gabriel-abrahao: It will, but I plan to add other, more disaggregated subtypes later when they are needed for harmonization, and thought better to already structure the code around that.

R/fullREMIND.R
```diff
@@ -110,14 +110,22 @@ fullREMIND <- function() {
...
-  calcOutput("GAINSEmi", subtype = "emission_factors",                          round = 5, file = "ef_gains.cs4r")
   calcOutput("GAINSEmi", subtype = "emissions_starting_values",                 round = 5, file = "f11_emiAPexsolve.cs4r")
-  calcOutput("EmissionFactors", subtype = "emission_factors", warnNA = FALSE,   round = 5, file = "f11_emiFacAP.cs4r")
-  calcOutput("EmissionFactorsFeedstocks",                                       round = 5, file = "f_nechem_emissionFactors.cs4r")
-  calcOutput("EmiLULUCFCountryAcc", subtype = "UNFCCC",                         round = 5, file = "p_EmiLULUCFCountryAcc.cs4r")
+  # calcOutput("GAINSEmi", subtype = "emissions",                                 round = 5, file = "emi_gains.cs4r")
+  # calcOutput("GAINSEmi", subtype = "emission_factors",                          round = 5, file = "ef_gains.cs4r")
+  # calcOutput("EmissionFactors", subtype = "emission_factors", warnNA = FALSE,   round = 5, file = "f11_emiFacAP.cs4r")
+
+  calcOutput("GAINS2025forREMIND", subtype = "emission_factors_remindsectors", warnNA = FALSE,                    round = 5, file = "f11_emiFacAP.cs4r")
```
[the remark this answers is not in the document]
gabriel-abrahao: This is definitely something that could be optimized, particularly for GAINSlegacy, as there's an extra dimension (ssp) that it doesn't really have and there's some repeated data because of that. But that's the most backward-compatible way to do it so far. I'd be glad to retire `GAINSlegacy` at some point, but given I've done all of it on the ScenarioMIP rush, I wouldn't be comfortable leaving people without that or their many project-specific scenarios anytime soon.

R/calcGAINS2025forREMIND.R
gabriel-abrahao: The negative weights are really weird, as that's supposed to be activity data, but they are present in the original data for some reason in OCEA_WHOL (Oceania) and TURK_WHOL (Turkey) for `Transformations_HLF` and `Transformations_NatGas`. I set those negative values to 0 for now, noting that emissions are also zero in those cases.

Regarding the zeros, I think that's intended, please correct me if my reasoning is wrong. We are aggregating emission factors (emissions per quantitiy of activity, say, tons of SO2 per PJ of energy generated). If the activity in a certain region is zero, it shouldn't count towards the aggregated emission factors. Those are supposed to have emission factors = 0 anyway, only there because NAs are a pain.

R/calcGAINS2025forREMIND.R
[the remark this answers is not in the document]
fbenke-pik: You don't get this warning per default in you  local environment, but we agreed to activate this warning on the Cluster, because it caused nasty bugs. To get it locally, add this to your .Rprofile
```
# increase warnings severity
options(warnPartialMatchDollar = TRUE)
```

R/fullREMIND.R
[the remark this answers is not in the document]
fbenke-pik: Ok, cool. If this is due to legacy code we cannot retire yet, then we cannot help it.

R/calcGAINS2025forREMIND.R
```diff
@@ -0,0 +1,177 @@
...
+#'
+#' @return Activity levels, emissions or emission factors
+#' @author Gabriel Abrahao
+#' @param subtype "emission_factors", "emissions","emissions_starting_values"
+#'
+#' @importFrom abind abind
+#' @importFrom utils head tail
+#' @importFrom magclass as.magpie
+#' @importFrom tidyr pivot_longer drop_na
+calcGAINS2025forREMIND <- function(subtype) {
```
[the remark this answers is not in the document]
fbenke-pik: Right. So assuming you are actually cool with this warning, you can deactivate it by adding 
`aggregationArguments <- list(zeroWeight = "allow")` the the list returned by your calc function

R/calcGAINS2025forREMIND.R
fbenke-pik: > please correct me if my reasoning is wrong

Sounds logical to me, so 'zeroWeight = "allow"` should be the way to go.

R/calcGAINS2025scenarios.R
```diff
@@ -0,0 +1,371 @@
...
+    weight = NULL, dim = 1, wdim = 1
+  )
+
+  # Reference activity relative change between 2025 and 2050
+  refract <- setYears(baseact[, 2050, ], NULL) / setYears(baseact[, 2025, ], NULL)
+  # Reference GDP relative change between 2025 and 2050
+  refrgdp <- setYears(gdpgains[, 2050, "SSP2"], NULL) / setYears(gdpgains[, 2025, "SSP2"], NULL)
+
+  # Estimate elasticities of the GDP-activities relationship
+  estela <- collapseDim(log(refract) / log(refrgdp))
```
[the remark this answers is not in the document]
fbenke-pik: Yes, it is in `calcGAINS2025scenarios`  `estela <- collapseDim(log(refract) / log(refrgdp))`
I added a comment in this PR as well.

DESCRIPTION
fbenke-pik: Ok, works for me, then.

PR #682, opened 2025-05-20
refactor calcEmiTarget and calcCapTarget

fbenke-pik:
Refactor `calcEmiTarget` and `calcCapTarget`:
- rewrite functions the madrat-way
- remove redundancies between NewClimate and UNFCCC_NDC
- add documentation 
- add TODOs for future refactorings and enhancements

This refactoring does not change the generate inputdata with one exception (CHA 2025 Emission Targets for NewClimate is dropped, but this is intentional)

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '467772398242'
+ValidationKey: '46781573'
 AcceptedWarnings:
 - Invalid URL: .*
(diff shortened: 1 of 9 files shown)


RahelMA (approved):
Thank you, Falk, for the comprehensive refactoring!

Note: The emission targets from NEWCLIMATE are currently not used in remind, therefore the change in the CHA emissions target will not affect any runs!

PR #688, opened 2025-06-10
Update F-gas exogenous trajectories

gabriel-abrahao:
Updates the default exogenous F-gas emissions inputs `f_emiFgas.cs4r` with data from new IMAGE2025 scenarios, in order to have Kigali-compliant scenarios in REMIND as mentioned by @RahelMA in the [development issues](https://github.com/remindmodel/development_issues/issues/476). The new implementation replaces the old one, changing the data that goes into the file instead of adding a new dimension, but I'd like to retain the old logic in `calcFGas` for a while just in case. The new data also still has the `SPA` dimension, although as only `SPA0` is [actually used in REMIND](https://github.com/gabriel-abrahao/remind/blob/a66522299be0c8bb345948e6edc7025fa0c1dacd/core/bounds.gms#L513), it's repeated in all entries.

The original scenarios we got from the IMAGE team are quite strange (see below), so we do some substantial processing on them before mapping them to our SSPs and RCPs. CP_GI is supposed to be a Kigali-compliant scenario, but only until 2040 for some reason. 

orig

The logic here is to use the short term development of CP_GI scenario in all RCPs up to RCP45, up until the actual scenario from IMAGE has lower emissions than CP_GI. And for RCP60, use the raw SSP baseline. This will basically give us (hopefully) Kigali-compatible scenarios for all the RCPs we really use (including RCP45 used in NPi) and RCP60 and "none" will be there for signalling a partial or total reversal of policies in case we want that. It's pretty crude, but we're working a bit blindly here.

IMAGE2025noint below is the new implementation as described above, and IMAGElegacy is what we had before. We get a lot less variation between scenarios, but that's coming from the data, and I personally don't think there's that much doubt in F-gas mitigation unless we are talking about a policy reversal scenario. For that case, we have "rcp60" (from the IMAGE 45I or 34I scenarios above) and "none" (from the baselines). The RCP60 here looks indeed very strange, but it's coming straight from the also strange weaker mitigation scenarios from IMAGE.

fgas_total

In default REMIND scenarios, the highest RCP we use is RCP45, so we should always be on the lowest cluster of scenarios. Comparing them with the previous implementation IMAGElegacy, the new setup converges to more or less the same values in the long run, but has substantially faster short term mitigation.


laurinks:
Good point, @merfort. I agree that this would make sense both for ScenarioMIP and given that we are mid 2025.

gabriel-abrahao:
> do you think it would make sense to also align the 2025 values for the F-Gases?
That does make sense, but do you know if we have reference values from ScenarioMIP? Right we use something the author of the function (someone I don't know called Gabriel Abrahao) [called EDGAR7, with data up to 2021](https://github.com/pik-piam/mrremind/blob/5d2a29cb16d7f081f98ec121371d28a6e8c5c860/R/readEDGAR7Fgases.R). The 2025 value is extrapolated with the 2010-2020 growth rate. Would it be helpful to start there?

> But just to be sure: this won't be relevant for ScenarioMIP since we will rely on the infilling process here anyways, right?
That's my understanding

gabriel-abrahao:
Either way, if we don't have anything better, I changed the default to the scenarios diverging only after 2025, using the extrapolation from EDGAR7 like mentioned above.
fgas_total

merfort (approved):

PR #690, opened 2025-06-13
manually adjust windon costs in calcDiffInvestCosts

fbenke-pik:


robertpietzcker:
The effect of this PR on CHA wind onshore costs can be seen in the blue line here (compared to red - old)
(although they also include the effect of this REMIND PR: https://github.com/remindmodel/remind/pull/2125
image

PR #695, opened 2025-06-25
Air pollution new scenario logic for ScenarioMIP

gabriel-abrahao:
This PR updates all new air pollution scenarios (`GAINS2025`) to use the June 24th data from Zig, and adds two more scenarios specific for ScenarioMIP. The legacy scenarios (`GAINSlegacy`) are untouched, and include some scenarios that seem scenario-specific (Namely `MFR_Transports`, `SLCF_building_transport`, `FLE_building_transport`, `GlobalEURO6`) that were not updated. Sector and Europe people (@robertpietzcker @Renato-Rodrigues ) please take note and let us know if we need to do something about them. If we don't, the old implementation will still work, but with old EFs. 

Now we have 6 AP scenarios in REMIND instead of the old 4, each with all SSP variations. The new ScenarioMIP approach from Zig does away with the CLE/SLE logic and basically maps scenarios to their SSPs, with small variations except for VLLO. So the idea here is to keep the CLE/SLE logic (for other non-ScenarioMIP applications) and add two scenarios, one that is specifically for ScenarioMIP VLLO and one that is the ScenarioMIP choice for each SSP in non-VLLO scenarios. So we have:

- CLE
  - Current legislation, IIASA-internal
- SLE
   - Stronger legislation, IIASA-internal
- MFR
   - Maximum Feasible Reduction. Not really intended for use, and more as a technical reference. Identical for all SSPs.
- VLE
   - Very strong legislation, ad-hoc, follows SLE until 2050 and then converges towards MFR in 2100
- SMIPperSSP
   - IIASA's choice for all ScenarioMIP scenarios except VLLO. The logic is a combination of CLE, SLE and MFR, with substantial variation across SSPs based on Marina's GCI. 
- SMIPVLLO
   - IIASA's choice for ScenarioMIP's VLLO, with a specific logic. Also varies by SSP, but only between SSP1 and SSP2. Other SSPs are a repetition of SSP2 for completeness sake. 

In general (but by no means always), the scenarios look like the example below for `pecoal.seel.pc.power`. The new CLE (SSP1 and SSP2) is generally a bit lower than the old one in `GAINSlegacy`, and varies just a bit with SSPs. `SMIPbySSP` naturally varies much more, and is generally much lower than CLE. `SSPVLLO` is close but lower than `SMIPbySSP.SSP1`. 
   
|   |  |
| ------------- | ------------- |
| image  | image  |

 I still have quite a lot of illness brain fog, so it would be great to have some eyes on those @nicobauer @laurinks @merfort @robertpietzcker . Here are plots for all the EFs that are used in the optimization (missing those that just go through `exoGAINS`)
 
[remefs_all.pdf](https://github.com/user-attachments/files/20908210/remefs_all.pdf)

Addresses [development_issues 591](https://github.com/remindmodel/development_issues/issues/591)

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '47125425'
+ValidationKey: '47233760'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '47088888'
+ValidationKey: '47215120'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'mrremind: MadRat REMIND Input Data Package'
-version: 0.232.4
-date-released: '2025-06-23'
+version: 0.233.0
+date-released: '2025-06-25'
 abstract: The mrremind packages contains data preprocessing for the REMIND model.
 authors:
 - family-names: Baumstark
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'mrremind: MadRat REMIND Input Data Package'
-version: 0.232.5
-date-released: '2025-06-30'
+version: 0.233.0
+date-released: '2025-07-03'
 abstract: The mrremind packages contains data preprocessing for the REMIND model.
 authors:
 - family-names: Baumstark
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: mrremind
 Title: MadRat REMIND Input Data Package
-Version: 0.232.5
-Date: 2025-06-30
+Version: 0.233.0
+Date: 2025-07-03
 Authors@R: c(
     person("Lavinia", "Baumstark", , "lavinia@pik-potsdam.de", role = c("aut", "cre")),
     person("Renato", "Rodrigues", role = "aut"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: mrremind
 Title: MadRat REMIND Input Data Package
-Version: 0.232.4
-Date: 2025-06-23
+Version: 0.233.0
+Date: 2025-06-25
 Authors@R: c(
     person("Lavinia", "Baumstark", , "lavinia@pik-potsdam.de", role = c("aut", "cre")),
     person("Renato", "Rodrigues", role = "aut"),
--- a/R/calcAirPollEmiRef.R
+++ b/R/calcAirPollEmiRef.R
@@ -1,22 +1,16 @@
 #' Calculate air pollutant emissions for a reference year, for use
 #' in combination with GAINS data at different sectoral aggregations
 #'
-#'
+#' @param subtype just "total" is supported
 #' @param baseyear year to take as a reference from CEDS, ignored for the EDGAR2005 LUC CO2 emissions
+#' @param outunits "Mt/yr" or "kt/yr"
+#' @param namesformat "GAINS2025" or "REMIND" or "REMINDexo", the standard to use for pollutant names
 #' @return magclass object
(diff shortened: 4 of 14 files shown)


gabriel-abrahao:
# (Almost) Final Update from Zig (2.7)
Since we're in a tight schedule, we need to merge this with a mixed version, where only the SMIPVLLO and SMIPbySSP scenarios are updated according to Zig's latest data. Furthermore, the new data is missing the detailed sector aggregation, so we're also using the old version of municipal waste EFs. 

### Data plots
Here's the final data that goes into REMIND's optimization:
[emefs4remind__all.pdf](https://github.com/user-attachments/files/21020315/emefs4remind__all.pdf)

And here's the full data, that goes into exoGAINS:
[coming soon]

### Background
Here's my mattermost message from earlier, outlining the rationale
Expanding on Laurin's second point above. Let's group the files we need from Zig in the following:

A: Activity data, needed for extensions and, crucially, for weighting. We have this now (1.7 1AM email exchange)
B: ScenarioMIP emission factors for aggregated sectors. This is the base of what we need for the SMIPbySSP and SMIPVLLO scenarios. This was shared today at 5am, I'm doing a thorough check but seems promising so far.
C: ScenarioMIP emission factors for detailed sectors. In fact, we only need this for Municipal Waste emissions (SMIPbySSP and SMIPVLLO).
D: The "For Lara" set he mentions in the email. These are the EF files (aggregated and detailed) needed for the basic scenarios CLE, SLE, MFR and VLE.
If we are in a tight schedule for the release and for ScenarioMIP, it would be quite trivial to change the input files and make a (nice and madrat-y) version of the input data that has the "final" data for all sectors except for Municipal Waste (which I recall doesn't emit a lot of AP) but only for the ScenarioMIP scenarios.

laurinks:
Just reviewed [rawefs_agg.pdf](https://github.com/user-attachments/files/21022593/rawefs_agg.pdf): Overall, the SMIPVLLO and SMIPbySSP scenarios look good from a bird's eye perspective when considering the transition from historic periods to future projections, the differentiation across SSPs and the difference between SMIPVLLO and SMIPbySSP.

A few notes (none of these is a showstopper, but could be pointed out to Zig at some point):
1. VOC from CHEM is zero in all regions for SMIPVLLO and SMIPbySSP, but was non-zero for CLE/SLE/MFR.
2. In some cases, SMIPVLLO jumps from zero to some non-zero value in 2050. For example, CHEM, NOx, CHA. I assume this is linked to missing regional data, and then VLLO transitioning to some global frontier value in 2050.
3. Power_Gen_Coal continues to show very low SO2 values in 2025 in CHA compared to EUR, i.e. @robertpietzcker's comment from [here](https://github.com/pik-piam/mrremind/pull/680#issuecomment-2943089486) still applies.
image

laurinks:
Review of [emefs4remind__all.pdf](https://github.com/user-attachments/files/21022598/emefs4remind__all.pdf):

- pecoal.sesofos.coaltr.indst: SMIPVLLO above SMIPbySSP for EUR
image

laurinks (approved):
Thanks so much, @gabriel-abrahao ! 

I reviewed the files and left comments. I only scrolled through the R code, but did not review that in detail. Maybe, @fbenke-pik, could have another look here.

gabriel-abrahao:
Thanks for the thorough reviews @laurinks ! 

> raweffs_agg.pdf
On your points here, all are valid, but it seems like all of that oddness comes directly from the original data. Given that those problems don't seem major, I think we can live with them for now.

> Review of [emefs4remind__all.pdf](https://github.com/user-attachments/files/21022598/emefs4remind__all.pdf):

That was a great catch! What happened was that the very old implementation (GAINSlegacy) assumes a certain way to weight the mapping between sectors that I just assumed was correct and reproduced it. But on a closer I don't really agree with it, and don't even quite understand exactly what `toolAggregate` was doing without a weight in that case, and it was causing very weird stuff to happen in several cases where several GAINS sectors were mapped to a single REMIND sector. I changed it to a weighted aggregation using the estimated emissions with GAINS activities, and now it's all making more sense. 

Several numbers changed a lot in the remind-sector EFs (and only there), but I'm a lot more confident in the new ones than the old, and I don't trust GAINSlegacy anymore for those cases, so when changing it made us farther to it, I think the newer is better. 
image

robertpietzcker:
@gabriel-abrahao 
great that you managed the update!

is there a follow-up process with Zig on the EUR-CHA stuff? 

I really would like to hear their explanation why today (2025), CHA has roughly 1/3 of the specific emissions for coal compared to EUR in pecoal.sesofos.coaltr.industry 

image

and also much lower NOX and SO2 emissions for coal electricity: 

image

robertpietzcker:
@laurinks yes, I fully agree. I just wanted to mention it so it doesn't get lost - thanks for opening the v3.6.0 issue!!

PR #698, opened 2025-07-03
Hopefully final AP emission factors update

gabriel-abrahao:
This updates all the non-legacy air pollution emission factors, including the CLE, SLE, etc. and the ScenarioMIP sets of scenarios (see #695 for implementation and scenario details) to use the data received from Zig on July 3rd, 

#### All sectors used in exoGAINS
General scenario comparison
[rawefs_agg.pdf](https://github.com/user-attachments/files/21040793/rawefs_agg.pdf)
Zooming in on ScenarioMIP scenarios and SSPs
[smiprawefs_agg.pdf](https://github.com/user-attachments/files/21040781/smiprawefs_agg.pdf)

#### REMIND sectors
There was a major overhaul in the sector aggregation weighting that I think is a lot more sensible now. Expect differences especially in sectors that had a many-to-many mapping, most notably `pecoal.sesofos.coaltr.indst`
[emefs4remind__all.pdf](https://github.com/user-attachments/files/21040731/emefs4remind__all.pdf)

### DIFF
--- a/R/calcGAINS2025scenarios.R
+++ b/R/calcGAINS2025scenarios.R
@@ -167,10 +167,11 @@ calcGAINS2025scenarios <- function(subtype, agglevel = "agg") {
 
   # Some sectors are only present in the historical data, having NAs in the
   # period in incle[,,"historical"] but simply being missing in inmfr and insle
-  # Here we bind that future period with NAs to the scenario, so that concatenation
+  # Here we bind that future period with NAs to the scenario (taking 2030 as
+  # reference), so that concatenation
   # between scenarios works and we can actually fill that data if needed after
   # the historical period is also concatenated.
-  padAbsentSectors <- function(magscen, incle) {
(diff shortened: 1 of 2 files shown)


LaviniaBaumstark (approved):

laurinks:
- I noted that CHEMBULK, AACID, and a few others are all zero. Is this an issue?
- In a few places, SSP4 looks weird. This is no surprise, I guess, as it copies SSP3 and there might be some effects from the underlying activities... --> I will take a not too add a warning on SSP4 to main.gms. Even though, I do not think we are going to have SSP4 REMIND anytime soon.
image

gabriel-abrahao:
> * I noted that CHEMBULK, AACID, and a few others are all zero. Is this an issue?
AACID was supposed to be like this, but I checked and indeed there was a time when CHEMBULK had data. I checked with Zig.

PR #699, opened 2025-07-07
Patching Air Pollution emission factors

gabriel-abrahao:
Motivated by Zig's email at the bottom, there were some additional patches and clarifications to the AP data we got from IIASA since the "final" update on July 3rd. This PR addresses them as below. I personally see very marginal benefits here so, given the time pressure and the potential of creating spillover problems as we can't take the time to check everything thoroughly, me and @laurinks are of the opinion this should be best left for merging after the ScenarioMIP release finalization.

#### New files
We got new files with data only for ScenarioMIP scenarios + historical (so no CLE, etc.) for sectors Transformation_Coal and Transformation_NatGas only. This PR adds them

#### Transformation_NatGas handling
Quoting Zig:
> recommend using this EF only as a relative change to your (or CEDS emissions of VOC from this section) since activity in GAINS is actually emissions.
We currently already do this for all the extraction sectors (which those are part of) in the exoGAINS script, so we're covered.

#### Faulty sectors
This PR sets to zero EFs for Transformation_HLF and Transformation_LLF as suggested.

#### exoGAINS files
All changes since the last version should be just here, in the Transformation sectors
[smiprawefs_agg.pdf](https://github.com/user-attachments/files/21102481/smiprawefs_agg.pdf)
[rawefs_agg.pdf](https://github.com/user-attachments/files/21102483/rawefs_agg.pdf)

#### REMIND sector files
No change expected since previous PR
[remefs_all.pdf](https://github.com/user-attachments/files/21102590/remefs_all.pdf)
[emefs4remind__all.pdf](https://github.com/user-attachments/files/21102628/emefs4remind__all.pdf)

## Email from Zig
Follow up on the Transformation sector headache. We have exchanged with few of you about it.

We have uploaded the four attached files to the 2-July folder. These are corrected EFs for:

- ‘Transformation_Coal’ sector - as indicated yesterday the values refer to coke production and you can use those also as relative change in EFs of course, and
- ‘Transformation_NatGas’ sector – here we recommend using this EF only as a relative change to your (or CEDS emissions of VOC from this section) since activity in GAINS is actually emissions. We have tried to refer this to natural gas production but were not successful yet. 

While working earlier on other issues, the Transformation sector got ‘infected’ and we have not noticed it.

The ‘Transformation_HLF_Refinery’ is fine and refers to crude oil throughput in refinery.

 

One more comment on the ‘Transformation_HLF’ and ‘Transformation_LLF’.  Currently, we recommend to ignore these since there might be a problem with the meaning of activity data here and this is what we struggled with producing unfortunately not consistent numbers, and some negative values. The July 2 set includes NOx and SO2 only and these are shall not be used. To my understanding these sources were made to allocate VOC and the key question is if this means that the VOC emissions from production, storage, and distribution of crude oil is lost and therefore it would be best to take the CEDS VOC value for historical years and grow with activity changes. While we loose here the potential changes due to the AP narratives, the uncertainties in emission factors are quite large and I think that change in activity data in SSPs and variants do account for key variability in the future. Exception might be the VLLO where one shall assume additionally reduction of losses from current values by about 60% by 2070.


merfort (approved):
Thanks, Gabriel!

fbenke-pik:
@gabriel-abrahao @laurinks Is this still relevant, or can it be closed?

gabriel-abrahao:
This was supeseded by #731

PR #700, opened 2025-07-08
read in 2024 UNFCCC data

fbenke-pik:
- This update adds an additional year 2022
- The read in logic has been revised to use a more robust logic that no longer depends on exact absolute position of data in all the sheets and parses UNFCCC identifiers instead. It also crashes when expected sheets or files are missing. Besides, the read in data uses variable names including the universal short codes of emission items (e.g. 1.A.1.b)

### Affected files

- Reference data in historical.mif via `calcUNFCCC` -> @pweigmann  (you might want to validate the GWP conversion factors you used when calculating GHG values, in 2023 they were a mix of AR4 and AR5; not sure countries reported in 2024)
- Emission Factors in input data `f_nechem_emissionFactors.cs4r` (mrindustry::calcEmissionFactorsFeedstocks) -> @JakobBD --> https://github.com/pik-piam/mrindustry/pull/56
- Historical LULUCF emissions following country accounting in input data `p_EmiLULUCFCountryAcc.cs4r` (mrremind::calcEmiLULUCFCountryAcc) -> @fschreyer 

**You can find comparison plots to get an overview of the changes here:**
https://cloud.pik-potsdam.de/index.php/f/24670641

important differences
- Higher Emissions in CAZ due to additional data in Australia that was not read in successfully with previous logic. The new values are higher, because AUS is now successfully included.
- LULUCF CO2 emissions from UNFCCC database are higher for REF, CAZ and USA now, they seem to have adjusted their reporting.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '47466900'
+ValidationKey: '47672100'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'mrremind: MadRat REMIND Input Data Package'
-version: 0.234.0
-date-released: '2025-07-16'
+version: 0.235.0
+date-released: '2025-07-17'
 abstract: The mrremind packages contains data preprocessing for the REMIND model.
 authors:
 - family-names: Baumstark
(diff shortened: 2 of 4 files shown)


JakobBD:
FYI @whitenightZhang and @leonieschweiger

fbenke-pik:
> > @pweigmann (you might want to validate the GWP conversion factors you used when calculating GHG values, in 2023 they were a mix of AR4 and AR5; not sure countries reported in 2024)
> 
> Great update! My understanding of the conversions is, that countries report their emissions as masses of the different gases and only in `calcUNFCCC` we convert to CO2eq. There I only see the AR5 conversion factors (28, 265) which are independent of the data source. Am I missing something? Where do we use the AR4 factors?

You are right, all good. I oversaw that we do not read in GHG values from UNFCCC, so we do not care which conversion factors they use.

PR #701, opened 2025-07-11
only report CEDS data from 1970 on in historical.mif

fbenke-pik:
This drastically reduces the number of mostly empty columns for years starting in 18th century in historical.mif


LaviniaBaumstark (approved):
yeah, thanks!

PR #703, opened 2025-07-14
Convert tax convergence data to expert guess

fbenke-pik:


laurinks:
Thanks for asking. As I have only been working on the refactoring of the REMIND code, but **not** on input data, I need to refer to @nicobauer for any questions on the actual values in the input data.

R/readExpertGuess.R
```diff
@@ -3,9 +3,12 @@
 #' Read-in data that are based on expert guess
 #'
 #' @md
-#' @param subtype Type of data that should be read.  One of
+#' @param subtype Type of data that should be read. One of
 #'   - `capacityFactorGlobal`: Global capacity factors for all REMIND technologies
 #'   - `capacityFactorRules`: Capacity factor rules for selected H12 regions and REMIND technologies
+#'   - `subConvergenceRollback`: Subsidy convergence level in rollback scenario
+#'   - `taxConvergence`: Tax convergence level in US$2005
+#'   - `taxConvergenceRollback`: Tax convergence level in rollback scenario
```
fbenke-pik, line 11: @nicobauer Can these explanations of the Expert Guesses be improved?

robertpietzcker:
I am unsure - I think from the technical side this is all fine, but it feels a bit weird to merge this and NOT fix the (likely) error of having a 0 convergence target level for EUR and JPN but 10 for all other regions.

so I would indeed wait for @nicobauer to return and say something about this question, so that we are sure not to propagate an error

fbenke-pik:
Makes sense. Let's wait for a comment by either @nicobauer or @laurinks

nicobauer:
This is indeed a bit strange. I have added the convergence for the roll-back scenarios.
The relevant fetaxscen is only scitched equal to "2" for three scenarios in the scenarioMIP config, that are not used. Correct, Laurin?
If I see correctly the relevant file is this one: /p/projects/rd3mod/inputdata/sources/ExpertGuess/tax_convergence_v1.0.csv
Here, the point is that EUR and USA do not appear. Thus the parameter value is equal to 0, while it is set to 10 explicitly for all other regions. Maybe I have not fully catched the line of info processing, though.

SMIPv06-VLLO-SSP1-PkPrice500-def
SMIPv06-VLLO-SSP2-PkPrice500-def
SMIPv06-VLLO-SSP3-PkPrice500-def

robertpietzcker:
> Couldn't it just be a fixed value in REMIND then? Or do we expect this value to vary per region in the future?

I had that thought as well, but wasn't sure about it. 

but yes - as this is the "convergence of taxes" scenario, it would also make sense to have only one tax level for all regions. 

so if it is easy for you to implement, why not just have one number and say "apply this number to all regions"

robertpietzcker:
> I have adjusted the input files differences now look as follows
> 
> ```
>        V2    V1    V3     x     y  diff   factor
> EUR   EUR y2050 fedie  0.70 11.68 10.98 16.68571
> EUR.1 EUR y2050 fepet  0.70 11.68 10.98 16.68571
> JPN   JPN y2050 fedie  0.00  9.48  9.48      Inf
> JPN.1 JPN y2050 fepet  0.00  9.48  9.48      Inf
> NEU   NEU y2050 fedie 11.06 11.83  0.77  1.06962
> NEU.1 NEU y2050 fepet 11.06 11.83  0.77  1.06962
> ``
> ```

ok, that is funny - we give it the same target level (10$), but because we use different $-conversions in different regions, they end up with a different tax level? 

could we change it such that the target levels are defined in $2017, so that there is no further unit conversion happening? 
(not important, but if it can be done easily, it might prevent future confusion why the regions end up with different tax levels in the convergence scenario)

robertpietzcker (approved):

PR #707, opened 2025-07-22
Ember update

pweigmann:
Update to 2025 release of Ember's [Yearly Electricity Data](https://ember-energy.org/data/yearly-electricity-data/). Contains 2024 data for most countries, but some are missing. Aggregation to regions is still performed as many variables appear to only miss negligible values. For 2023 most gaps seem to now be filled.

This documents gives an overview where data has changed since the last release:
https://cloud.pik-potsdam.de/index.php/f/24786713

Of interest might be the changes to historical solar capacities in REF and CAZ regions:
image


robertpietzcker:
This will also update emissions in the power sector  in historical.mif, right?

PR #708, opened 2025-07-28
remove transport related madrat warnings

fbenke-pik:
This removes zero weight aggregation warnings. 

Background: in `calcEDGETransport`, some weight manipulation was done, using the current regional resolution as a determining factor in the calc function. This is bad practice and led to problems when using 21 regional resolution due to unwanted caching. The solution is moving the logic to a custom aggregation function which can access the current regional resolution dynamically and independent of the caching.

Note that this bugfix introduces values for EU21 regions that were dropped before due to zero weights for these regions (e.g. FRA), but this is a correction and they should have been there in the first place.


R/calcEDGETransport.R
```diff
@@ -30,6 +30,7 @@ calcEDGETransport <- function(subtype) {
       weight <- as.magpie(weight)
       unit <- "2017US$/(p|t)km"
       description <- "Capital cost (purchase) per energy service demand on CES level."
+      aggregationArguments <- list(zeroWeight = "allow")
```
fbenke-pik, line 33: @alexkoberle @johannah-pik 

The subtypes "f35_esCapCost"  and "f34_fe2es"  throw a warning `~ Weight sum is 0, so cannot normalize and will return 0 for some aggregation targets. This changes the total sum of the magpie object! If this is really intended set zeroWeight = "allow", or "setNA" to return NA.`

This happens whenever the chosen weight is zero for one of the aggregation regions - resulting in the values of that region to be zero due to multiplication with 0. Sometimes this is a problem, sometimes it isn't. 

Could you take a look and let me know if this is fine?

ahagen-pik (approved):

ahagen-pik:
Thank you, @fbenke-pik for spotting and solving this bug. 

I think it might explain why we had such big differences in aggregated regions in a H12 and EU21 run.
image

This was already partly solved be the new edgeTransport version 3.0.0 ("startyear" in the plot) because (dis)aggregation is now included in iterativeEdgeTransport an it no longer depends on mrremind input data. This fix was 'by accident' and without knowledge of the bug.

This PR now consistently fixes the issue for input data.
Great work!

PR #713, opened 2025-07-30
Climate trace

pweigmann:
Importing a new data source for historical emissions: [ClimateTrace](https://climatetrace.org)

They use remote sensing and a range of other methodologies to come up with their own estimates of emissions and rely on EDGAR/FAOSTAT in a few categories. The results deviate from more "traditional" sources and are overall higher and more up to date (monthly releases).

@robertpietzcker could you take a look at the mapping? I managed to fill out everything but am not completely sure about it. The biggest problem seems to me that they don't differentiate between ``Energy|Demand|Industry`` and ``Industrial Processes`` emissions, so I mapped directly to "Industry" being the sum of the two. Their sectors are defined here: https://climatetrace.org/sectors

In a second step I would create some plots to compare to our other data sources and invite people to discuss the value of the source.


R/convertClimateTrace.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' @md
+#' @param x A [`magpie`][magclass::magclass] object returned from
+#'          [`readClimateTrace()`].
+#'
+#' @return A [`magpie`][magclass::magclass] object.
+#'
+#' @author Pascal Weigmann
+#'
+#' @export
```
fbenke-pik, line 11: I think export does not make sense for madrat functions, as they should never be called directly, but rather via readSource or calcOutput.

pweigmann: Indeed, missed that after copying from another file!

inst/extdata/sectoral/mappingClimateTrace.csv
```diff
@@ -0,0 +1,69 @@
...
+forest-land-clearing;Land-Use Change;
+forest-land-degradation;Land-Use Change;
+forest-land-fires;Land-Use Change;
+glass;Industry;
+heat-plants;Energy|Supply|Heat;
+incineration-and-open-burning-of-waste;Waste;
+industrial-wastewater-treatment-and-discharge;Waste;
+international-aviation;Transport|Pass|Aviation|International|Demand;
+international-shipping;Transport|Freight|International Shipping|Demand;
+iron-and-steel;Energy|Demand|Industry|Steel;
```
robertpietzcker, line 29: this seems to be the only industry subsector that is split out explicitly. Is this still added to the industry total somewhere?

robertpietzcker:
Thanks for the explanations, @pweigmann !
great if they have bunkers on national level, although it will be interesting to see how their emissions split matches the IEAs FE split :-)

I guess adding it now is a good first step, and then the next steps might be to add the disaggregations, in exchange with the sector teams. 

or - would it hurt to have the transport subsectors directly read in close to the original detail level, so into the following variables: 
- Emi|CO2|Energy|Demand|Transport|Rail
- Emi|CO2|Energy|Demand|Transport|Road
- Emi|CO2|Energy|Demand|Transport|Pass|Domestic Aviation
- Emi|CO2|Energy|Demand|Transport|Freight|Domestic Shipping

and then aggregated first to Emi|CO2|w/o Bunkers|Energy|Demand|Transport and then Emi|CO2|w/ Bunkers|Energy|Demand|Transport?

That would mean changing the transport mappings as follows: (I guess you need to remove the "Emi|Co2", but I kept it here to show how the final variable looks like)

domestic-aviation;Emi|CO2|Energy|Demand|Transport|Pass|Domestic Aviation;
domestic-shipping;Emi|CO2|Energy|Demand|Transport|Freight|Domestic Shipping;
international-aviation;Emi|CO2|Energy|Demand|Transport|Bunkers|Pass|International Aviation;
international-shipping;Emi|CO2|Energy|Demand|Transport|Bunkers|Freight|International Shipping;
non-broadcasting-vessels;Emi|CO2|Energy|Demand|Transport|Freight|Domestic Shipping;
other-transport;Emi|CO2|Energy|Demand|Transport|Road;
railways;Emi|CO2|Energy|Demand|Transport|Rail;
road-transportation;Emi|CO2|Energy|Demand|Transport|Road;


and then add
Emi|CO2|Energy|Demand|Transport|International Bunkers = Emi|CO2|Energy|Demand|Transport|Bunkers|Pass|International Aviation + Emi|CO2|Energy|Demand|Transport|Bunkers|Freight|International Shipping

pweigmann:
> Thanks for the explanations, @pweigmann ! great if they have bunkers on national level, although it will be interesting to see how their emissions split matches the IEAs FE split :-)

Sounds good! Wouldn't hurt at all, I'll do the changes.

pweigmann:
> domestic-aviation;Emi|CO2|Energy|Demand|Transport|Pass|Domestic Aviation;
> domestic-shipping;Emi|CO2|Energy|Demand|Transport|Freight|Domestic Shipping;
> international-aviation;Emi|CO2|Energy|Demand|Transport|Bunkers|Pass|International Aviation;
> international-shipping;Emi|CO2|Energy|Demand|Transport|Bunkers|Freight|International Shipping;
> non-broadcasting-vessels;Emi|CO2|Energy|Demand|Transport|Freight|Domestic Shipping;
> other-transport;Emi|CO2|Energy|Demand|Transport|Road;
> railways;Emi|CO2|Energy|Demand|Transport|Rail;
> road-transportation;Emi|CO2|Energy|Demand|Transport|Road;

Altough these standard variables are now **including** bunkers, right? So in this case they should be ``Emi|CO2|w/o Bunkers|...`` right?

pweigmann:
@robertpietzcker last questions:
- I realized the ``non-broadcasting-vessels`` are described as "Air emissions from ocean-going vessels...", so rather international than domestic?
- and ``iron-mining`` etc. go into ``Industry`` then?

PR #716, opened 2025-08-01
IRENA Update to 2025 release

pweigmann:
One more year of capacities and power generation from IRENA.

Data is used in the following functions:

1. calcCapacity -> capacity, changing 2023 to 2024 ok like this @lecfab?
2. calcCapacityFactorHist -> cap + gen, no change uses 2015 only
3. calcDspvShare -> cap, no changes (except formatting), uses old data only
4. calcIRENA -> no changes needed
5. calcPotentialGeothermal -> cap, uses 2020 value only
6. calcPotentialHydro -> cap + gen, no change uses 2015 only
7. convertNewClimate -> cap + gen, uses 2015 only, should be fine, right @RahelMA?
8. convertREN21 -> cap + gen, uses 2015 only
9. readUNFCCC_NDC -> cap + gen, uses 2015 only, should be fine, right @RahelMA?

Doing a test input-data generation with the updated package and won't merge until it finished successfully.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '48537300'
+ValidationKey: '48559992'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'mrremind: MadRat REMIND Input Data Package'
-version: 0.239.1
-date-released: '2025-07-31'
+version: 0.239.2
+date-released: '2025-08-01'
 abstract: The mrremind packages contains data preprocessing for the REMIND model.
 authors:
 - family-names: Baumstark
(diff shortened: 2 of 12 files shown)


fbenke-pik (approved):

fbenke-pik:
> remove the renaming, as I think it is useless following @fbenke-pik's https://github.com/pik-piam/mrremind/pull/675#discussion_r2068182222.

If you remove the temporal mapping, then you will get the unaggregated values as input data, i.e. values for 2023 and 2024, but no 2025 value.

> remove the 2023 values so that the averaging only takes the 2024 value (if I understand the temporalmapping correctly).

I think this is bad practice, as you drop data for non-obvious reasons. If you don't like the result of temporal aggregation, drop it altogether rather than tweak data to get results you like better.

lecfab:
> If you remove the temporal mapping

I wasnt suggesting that: just removing the renaming of "2024 to 2025". That way, there are values only for 2023 and 2024, and the temporal mapping still takes their average to output a 2025 value.

> If you don't like the result of temporal aggregation, drop it altogether 

We like it for anytime before 2022, but it gets tricky when we only have incomplete value for the 2025 timestep (that's why I had a dirty way around [initially](https://github.com/pik-piam/mrremind/pull/675#discussion_r2068182222)).

PR #719, opened 2025-08-06
update temporal mapping for tech capacities

fbenke-pik:
For `pm_histCap.cs3r`, only use the year 2024 for the period 2025.


lecfab (approved):
Very elegant solution thanks!

PR #721, opened 2025-08-08
Update IAEA PRIS nuclear project database

pweigmann:
Updated the numbers from the IAEA PRIS Website that is being used to arrive at near-term estimates for nuclear capacities in 2030. For more details see the documentation here: https://github.com/pik-piam/mrremind/discussions/540

Also removes some code reading in data from the Global Energy Monitor as it currently isn't useful.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '48754706'
+ValidationKey: '48779816'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'mrremind: MadRat REMIND Input Data Package'
-version: 0.240.1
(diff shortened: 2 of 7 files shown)


robertpietzcker:
@pweigmann great! 
did you compare the outputs?

I remember that there are some calculations based on the data to come up with the REMIND input data that likely need to be adjusted (because more recent data is available, so the "remaining growth formula" needs to be updated. 

did you do that already, or is that the next step?

pweigmann:
> @pweigmann great! did you compare the outputs?
> 
> I remember that there are some calculations based on the data to come up with the REMIND input data that likely need to be adjusted (because more recent data is available, so the "remaining growth formula" needs to be updated.
> 
> did you do that already, or is that the next step?

There were only very minor changes with this update. I didn't create any plots but you can see the before-after in the two files in ``/p/projects/rd3mod/inputdata/sources/IAEA_PRIS``.

It doesn't affect REMIND though, as there are two separate functions: ``readIAEA`` for input-data (via ``calcCapacityNuclear``) and ``readIAEA_PRIS`` for near-term thresholds (via ``calcProjectPipelines``). It's on my list of nice-to-have things to merge these two, however currently not as a priority.

PR #723, opened 2025-08-18
remove base prefix

fbenke-pik:
This removes the base prefix from pmin and pmax, as these functions are being refactored in magclass
https://github.com/pik-piam/magclass/pull/185


pascal-sauer:
Some changes unrelated to pmin are happening in readUBA with this PR, is this intended?

fbenke-pik:
> Some changes unrelated to pmin are happening in readUBA with this PR, is this intended?

No, it got included by accident, but should not be a problem.

PR #727, opened 2025-08-21
Adjust hydro potential to avoid ECE infeasibility

Renato-Rodrigues:
- Increase the hydro potential to avoid ECE infeasibility in recent runs

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '48930560'
+ValidationKey: '48953289'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


LaviniaBaumstark (approved):
thanks

dbantje (approved):
Great, thank you @Renato-Rodrigues ! No comments from my side, I'll retry my runs probably only once I'm back from vacations in September.

Renato-Rodrigues:
> Thanks for the fix. As this caused issues several times, do you think we should state in mrremind that the maxprod _has to_ be above NDC targets, (assuming that countries define feasible targets)?

There is already a comment in the code regarding this:
https://github.com/pik-piam/mrremind/blob/34222d49f1be16bcb8ddea06bf88e68ac99e717c/R/calcPotentialHydro.R#L206

Ideally, hydro potentials should be reworked using country-level data. Atreya and I started some work on this a few years ago, but since the impact on model results was limited, it was deprioritized and has remained frozen since then.

Alternatively, one could add code that checks NDC targets and ensures the potentials are consistent with them. However, I am not directly involved in defining NDC targets.

PR #730, opened 2025-08-26
move industry-related subtypes in calcIO to mrindustry

fbenke-pik:


JakobBD (approved):
Very helpful, thanks!

PR #738, opened 2025-09-16
remove bound checks from calcEmiTarget

fbenke-pik:


fbenke-pik:
Closing, as this is a useful warning and the underlying [issue](https://github.com/pik-piam/mrremind/issues/740) should be handled

RahelMA:
You are right, as we have `ghgFactorMin` as well as `ghgFactorMax` this warning is redundant and can be removed. Thank you, Falk!

PR #742, opened 2025-09-18
read in renewable share targets from NewClimate protocol and implement method to aggregate them to REMIND regions

fschreyer:


R/fullREMIND.R
```diff
@@ -12,7 +12,6 @@
 #' }
 #'
 fullREMIND <- function() {
-
```
fbenke-pik, line 15: I think most of the formatting in this file is on purpose. I personally don't the formatting change, but @LaviniaBaumstark should approve this as well.

fschreyer: Oh, that actually was not on purpose. I only wanted to do it on the other files. Should I revert this?

R/convertNewClimate.R
```diff
@@ -353,6 +353,26 @@ convertNewClimate <- function(x, subtype, subset) { # nolint: object_name_linter
...
+      "POL", "CZE", "ROU", "BGR", "HUN", "SVK", "LTU", "EST", "SVN",
+      "LVA", "DEU", "FRA", "ITA", "ESP", "NLD", "BEL", "GRC", "AUT",
+      "PRT", "FIN", "SWE", "IRL", "DNK", "LUX", "CYP", "MLT", "JEY",
+      "FRO", "GIB", "GGY", "IMN", "HRV"
+    )
+
+    # add missing countries with NA
+    x <- toolCountryFill(x)
+    # assign share EU values to EU countries
+    x[EUR_NPi_countries, , ] <- x_EU
```
fbenke-pik, line 373: As far as i can see, "EU" is not removed from `x`. Does this throw a warning, as "EU" is not part of the 249 ISO countries?

R/readNewClimate.R
```diff
@@ -59,6 +57,30 @@ readNewClimate <- function(subtype, subset) {
       toolProcessClimateTargetDatabase(database = "NewClimate", subtype = subtype)
 
     x <- as.magpie(input, spatial = "ISO_Code", temporal = "Target_Year")
+    # read in energy share targets from policy modeling protocol
+  } else if (grepl("RenShareTargets", subtype, fixed = TRUE)) {
+    data <- readxl::read_excel(
```
fbenke-pik, line 62: did you update the source file without updating the file version? better increase the file version when you add content to the source to ensure backwards compatability

R/convertNewClimate.R
fbenke-pik: Edit: please fix 
```
~~ WARNING: Data for following unknown country codes removed: EU
~~ NOTE:  - toolCountryFill set missing values for IMPORTANT countries to NA:
...
```

They just create span in the input data log

R/calcRenShareTargets.R
```diff
@@ -0,0 +1,352 @@
...
+      # get total
+      xHistTotal[, , targetType] <- xHistTotalVar
+    }
+
+    return(list(xHistShare,xHistTotal))
+  }
+
+
+  # function to extrapolate target years which are not REMIND timesteps to REMIND timesteps
+  extrapolateTargetYears <- function(x, xHist, targetYears) {
```
fbenke-pik, line 165: if you decide to use nested functions, i recommend using a leading dot (e.g. `.extrapolateTargetYears` )  to make it clear to the reader that this is an internal function not used anywhere else

R/calcRenShareTargets.R
```diff
@@ -0,0 +1,352 @@
...
+  # interpolate and extrapolate renewable share targets for countries with targets
+  # assume that countries with share targets reach their share linearly from 2020 shares to the target year
+  # assume that after target year, renewable share targets stay constant
+  # assume that countries without target keep their historic share in the future
+  x_intp <- x
+  x_intp[,"y2020",] <- xHistShare
+  x_intp <- as.quitte(x_intp) %>%
+              mutate( "variable" = .data$data) %>%
+              select(-.data$data) %>%
+              interpolate_missing_periods(expand.values = T) %>%
```
fbenke-pik, line 307: FYI: You do not need to convert to quitte just to do interpolation. You could use `toolFillYears` as well to do it directly on the magclass object. However, as the NA years per time series vary in `x_intp`, you would still have to iterate over regions and third dimension (like here https://github.com/pik-piam/mrremind/blob/master/R/calcGAINS2025.R#L217-L224), so not sure if this is more elegant.

R/toolCalcEnergyProj.R
```diff
@@ -0,0 +1,142 @@
...
+    } else {
+      # only get 2020 year
+      xHist <- collapseDim(xHist[, "y2020", ])
+    }
+
+    return(xHist)
+  }
+
+  # function to calculate FE trend on level of REMIND regions based on EDGE data
+  calcFeTrend <- function(subtype, subset, years) {
```
fbenke-pik, line 55: I would not call it like this, as it could be confused with a madrat calc function, which it is not.

R/toolCalcEnergyProj.R
```diff
@@ -0,0 +1,142 @@
+#' Calculate energy projections on country-level based on EDGE models outputs per country.
+#' These energy projections are used in the input data preparation for aggregating country-specific data to REMIND regions.
+#' They are a country-level proxy of the final energy demand trajectories on the level of REMIND regions provided by the EDGE models.
+#'
+#' @author Felix Schreyer
+#' @param subtype "FE" (Total final energy consumption), "SE|Electricity" (SE electricity generation)
+#' @param subset String, designating the GDP scenarios to use
```
fbenke-pik, line 7: Pleas apply auto format and auto indent to fix trivial linter errors.

R/calcRenShareTargets.R
```diff
@@ -0,0 +1,352 @@
...
+        # calculate annual change rate between 2020 historical share and target year share
+        ShareChangeRate <- (x[, targetYear, ] - xHist) / (targetYear - 2020)
+        # next five-year time step
+        NextTimeStep <- targetYear - (targetYear %% 5) + 5
+        # linearly extrapolate energy share target to next five-year time step
+        x[, targetYear, ] <- x[, targetYear, ] + (5 - (targetYear %% 5)) * ShareChangeRate[, targetYear, ]
+        # if there is already a value for the next REMIND time step, use the higher value
+        # if not, assign extrapolated value to next REMIND time step
+        if (NextTimeStep %in% getYears(x, as.integer = TRUE)) {
+          x_target[, NextTimeStep, ] <- base::pmax(x[, targetYear, ], x[, NextTimeStep, ], na.rm = TRUE)
```
fbenke-pik, line 182: Should no longer be necessary to use base:: prefix for pmax

R/toolCalcEnergyProj.R
```diff
@@ -0,0 +1,142 @@
...
+        "feelwlth_otherInd", "feelhth_otherInd"
+      )
+      input.fe.trans <- c("feelt")
+    }
+
+    # common years across EDGE outputs
+    common.years <- intersect(getYears(FEBuildIndustry), getYears(FETransport))
+
+    # sum FE demand across sectors per REMIND region
+    FeTotal <- dimSums(FEBuildIndustry[, common.years, input.fe.build], dim = 3) +
```
fbenke-pik, line 92: Are you sure that you cannot get the total FE demand another way. You calculation is expensive and not robust in terms of caching due to the sources used. Besides, is seems redundant to me, as I would expect that we already have madrat functions that calculate this?

R/toolCalcEnergyProj.R
```diff
@@ -0,0 +1,142 @@
...
+      dimSums(FEBuildIndustry[, common.years, input.fe.indst], dim = 3) +
+      dimSums(FETransport[, common.years, input.fe.trans], dim = 3) * 31.536 # convert TWa/yr to EJ/yr
+    # calculate FE trend of target year with respect to 2020
+    FeTrend <- collapseDim(FeTotal[, years, ] / FeTotal[, "y2020", ])
+    # get H12 REMIND regionmapping
+    mapping <- toolGetMapping(type = "regional", name = "regionmappingH12.csv", where = "mappingfolder")
+    # assign REMIND FE region trend to all countries within this region
+    FeTrend <- toolAggregate(FeTrend, rel = mapping)
```
fbenke-pik, line 100: This crashes for other regional resolutions than H12. At this point you should not assume regions and operate on iso country level for your calculations.

PR #796, opened 2026-03-12
use IEA data postprocessing in calcIO and update variable mappings

fbenke-pik:


fbenke-pik:
Closed, because files have been moved to other packages

PR #804, opened 2026-03-30
Change NDC target calculation to provide absolute emissions targets to REMIND and remove country-level target filtering

fschreyer:
This PR changes our approach of implementing NDC emissions targets. Instead of providing relative emissions targets per region that are subsequently multiplied by REMIND emissions from historical periods, the new approach now calculates absolute emissions targets that are directly used in REMIND. This step is reasonable by now because we consistently include all emissions sources (incl. LULUCF) in our target calculation so we do not need to rely on relative targets anymore.  Overall, it simplifies our target calculation. Moreover, we do not filter out too ambitious/unambitious NDC targets anymore since we have constraints on the co2 price implemented in REMIND (co2 price has upper limit in near-term and lower limit from NPI co2 price). Changes in summary:

* switch from relative to absolute NDC emissions targets provided to REMIND
* include all NDC emissions targets (not filter out too ambitious/unambitious targets)
* clean up target calculation and rename some files/functions

Corresponding REMIND PR that adapts to the new approach will follow soon. 

Input data is here: ``/p/tmp/schreyer/Modeling/pik-piam/log-2026-03-30-7798148.out``

Change is implemented in REMIND via: https://github.com/remindmodel/remind/pull/2328


R/toolCalcGhgTarget.R
```diff
@@ -161,75 +135,122 @@ toolCalcGhgFactor <- function(x, subtype, subset) {
...
-  ghgFactor <- new.magpie(
+  # 3. Deal with Targets for EU countries ----
+
+  # disaggregate (relative) emissions targets defined for EU(27) to country-level
+  # get H12 regionmapping
+  regionmapping <- toolGetMapping("regionmappingH12.csv", where = "mappingfolder", type = "regional")
+
+  # apply EU emissions reduction goals uniformly to all countries in EUR region (EU28)
+  # for which emissions goals are not defined in the input data
+  EUR_NDC_countries <- regionmapping %>%
```
RahelMA, line 181: Ok, you now take the EU regions from the mapping directly instead of defining them manually. Have you checked if GBR is included? It has a separate NDC target ...

R/toolCalcGhgTarget.R
```diff
@@ -161,75 +135,122 @@ toolCalcGhgFactor <- function(x, subtype, subset) {
...
-  AbsTarget <- ghgFactor
 
-  # for each country and year, calculate calculate GHG factor
+  # for each country and year, calculate calculate absolute NDC emissions target in MtCO2eq/yr for total GHG emissions excl. bunkers and excl. LULUCF
   for (regi in getItems(reductionData, dim = 1)) {
     for (year in getYears(reductionData, as.integer = TRUE)) {
       if (!is.na(reductionData[regi, year, conditional][1])) {
+        # determine target year to define NDC emissions goal for, round to nearest REMIND time step
         y <- if (year < 2060) ceiling((year - 1) / 5) * 5 else ceiling((year - 2) / 10) * 10
+        # calculate absolute NDC emissions targets per country
```
RahelMA, line 248: Could you add a description of the assumption for the EU target calculation?

e.g. we use the historical emission by EU country; therefore, we assume that each country reduces emissions by the same percentage... Its nothing new; I just think it has always not been documented ...

RahelMA (approved):
Thank you, Felix! This is great work! I just have minor comments....

R/toolCalcGhgTarget.R
fschreyer: UK only gets EU emissions target if UK target is not defined. See [here](https://github.com/fschreyer/mrremind/blob/6a94b50c2a6d1189a583ea886cde414bb0d9a776/R/toolCalcGhgTarget.R#L183).

R/toolCalcGhgTarget.R
fschreyer: addressed

R/calcEmiTargetReference.R
```diff
@@ -79,16 +79,6 @@ calcEmiTargetReference <- function() {
   tmp[is.na(tmp)] <- iiasaHist[is.na(tmp)]
   out[, getYears(tmp), getNames(tmp)] <- tmp[, , ]
 
-  out <- add_columns(out, addnm = c("y2030", "y2035"), dim = 2)
-
-  # add forecast for 2030 from IIASA
-  fc2030 <- readSource("IIASALanduse", subtype = "forecast2030")
```
fbenke-pik, line 85: @fschreyer 
The code for these two subtypes of `IIASALanduse` can be deleted from readIIASALanduse as well, as they were only used here. Please clean this up as well.

fschreyer: It is now shifted to [toolcalcGhgTarget](https://github.com/fschreyer/mrremind/blob/da560afcb4fae0ad4f7bb07fba5e63e9e2a9f111/R/toolCalcGhgTarget.R#L209).

fbenke-pik: Ah, cool. Nvm then :)

PR #807, opened 2026-04-22
Add regional fossil capacity / power generation variables from IEA WEO2025 to historical mif

fschreyer:
This is such that the IEA WEO2025 data shows up in the SE|Electricity|Coal/Gas and Cap|Electricity|Coal/Gas plots.


fbenke-pik (approved):
Thanks :)

PR #810, opened 2026-05-07
bugfix NDC targets: emissions data used for aggregation weights for India/China were accidentally set to 0

fschreyer:
Emissions data used for aggregation weights for India/China were accidentally set to 0 as we overwrote with UNFCCC data for all years. Use CEDS data for this again and only replace 2005 values by UNFCCC data as these are the India/China target reference years. 

The India/China targets are as now what I expected from back-of-the-envelope calculation:

```
HPC2024 - schreyer@login03:/p/tmp/schreyer/Modeling/remind/Dev1/output/SSP2-NDC_2026-05-04_15.39.36> dumpgdx fulldata.gdx p45_CO2eqwoLU_goal "(CHA|IND)"
2030  CHA  14386.8684505262
2030  IND  4526.6211
2035  CHA  15165.5419266043
2035  IND  5362.2137
```

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '54561598'
+ValidationKey: '54619320'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'mrremind: MadRat REMIND Input Data Package'
-version: 0.265.3
-date-released: '2026-04-23'
+version: 0.265.4
+date-released: '2026-05-07'
 abstract: The mrremind packages contains data preprocessing for the REMIND model.
 authors:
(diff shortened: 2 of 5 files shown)


RahelMA (changes requested):
Thank you, Felix, for spotting this!

Can this not also be the case for other countries? If UNFCCC values are 0 or NA we should rely on CEDS but that holds for all countries. I think a more generic correction would be useful.

fschreyer:
I see your point. At the moment, this is not a problem because there are no NAs / 0 in the UNFCCC data for the regions we use (see below). Could be written more robust in the future. But, I would postpone this to the refactoring. 

```
Browse[1]>   out[unfcccReg, , ] <- ghgUNFCCC[unfcccReg, , ]
Browse[1]> check <- as.data.frame(ghgUNFCCC[unfcccReg, , ])
Browse[1]> check %>% filter(is.na(Value))
[1] Cell   Region Year   Data1  Value 
<0 rows> (or 0-length row.names)
Browse[1]> check %>% filter(Value == 0)
[1] Cell   Region Year   Data1  Value 
<0 rows> (or 0-length row.names)
```

RahelMA (approved):

PR #812, opened 2026-05-19
enhance IEA WEO mapping

fbenke-pik:
This PR introduces an enhanced IEA WEO mapping that allows for different mappings on ISO level compared to global level.


robertpietzcker (approved):
very helpful -thanks!

PR #813, opened 2026-05-20
Emission data from CEDS as used in CMIP7

pweigmann:
This adds emissions for multiple species and 13 sectors to the historical.mif together with the corresponding [mrcommons PR](https://github.com/pik-piam/mrcommons/pull/194). It uses a new mapping and a new tool function to handle global and country data flexibly.

Data is taken from three files, to complement country-level data with aggregated GHG/global data, which are kept as reference, e.g. to see how close our aggregation is after mapping to REMIND variables. Those references are

- `Emi|CMIP7|CO2|AFOLU` (does not include burning variables)
- `Emi|CMIP7|CO2|w/o AFOLU` (originally called Energy and Industrial Processes, which basically means w/o AFOLU in CMIP7 defintions)
- `Emi|CMIP7|GHG AR6GWP100`
- `Emi|CMIP7|Kyoto GHG AR6GWP100`

Currently, the following aggregations are calculated:
- `Emi|<species>`  (global only)
- `Emi|<species>|Energy and Industrial Processes`   (global only)
- `Emi|<species>|w/ Bunkers|Energy and Industrial Processes`   (global only, same as directly above)
- `Emi|<species>|w/o Bunkers|Energy and Industrial Processes`
- `Emi|<species>|w/o AFOLU` (global only)
- `Emi|<species>` (global only, no CO2)
- `Emi|GHG|<variables that exist for CO2, N2O and CH4>`

Note: CO2 totals do not include burning variables, other species do!

@laurinks let me know if you need others

@fbenke-pik I adjusted the custom aggregation function a tiny bit by introducing the additional argument `removeAllAgg`. The way I see it, we have two potential use cases for the function:
- as used in IEA WEO, where global data is shipped for all variables and preferred because countries are incomplete
- as used here, where some variables only have global data, other are complete sets of countries

In the first case we would want to overwrite/remove the standard country aggregation and use the global values (`removeAllAgg = TRUE`), in the second case we would want to keep both (`removeAllAgg = FALSE`). I am unsure if there might be other cases that are not covered, but if you think it works for now we could replace the custom aggregation with the new tool.

Source: https://zenodo.org/records/17845154

The output as it would appear in the historical.mif (H12 aggregation) can be seen in the attached file.

~~[historical_CEDS.xlsx](https://github.com/user-attachments/files/28066353/historical_CEDS.xlsx)~~
Latest: [historical_CEDS.xlsx](https://github.com/user-attachments/files/28145343/historical_CEDS.xlsx)


R/toolAggregateWithoutGlobal.R
```diff
@@ -0,0 +1,47 @@
...
+#' @param rel aggregation mapping for ``toolAggregate``
+#' @param to aggregation target for ``toolAggregate``
+#' @param removeAllAgg decide whether to exclude all aggregated data or keep
+#' those (variables/periods) that are not overwritten by data from glo object
+#' @author Falk Benke, Pascal Weigmann
+#' @returns magclass object
+#'
+#' @export
+
+toolAggregateWithoutGlobal <- function(x, glo, rel,
```
fbenke-pik: Could you please also replace .customAggregate in calcIeaWorldEnergyOutlook with this tool function?

R/calcEmissions.R
```diff
@@ -993,6 +1107,8 @@ calcEmissions <- function(datasource = "CEDS16") {
     x = tmp,
     weight = NULL,
     unit = "Mt",
+    aggregationFunction = toolAggregateWithoutGlobal,
```
fbenke-pik: Looks to me like this return statement applies to all subtypes, but only your new one actually sets `global`. Wouldn't this crash for other subtypes as is?

pweigmann: it would indeed, fixing

fbenke-pik: kk, maybe use the standard aggregationFunction for other subtypes as well, just to be safe

laurinks:
Question regarding the source: Is not this one the source we are relying on to get the full picture of emissions used in ScenarioMIP for CMIP7, which goes beyond CEDS?

https://zenodo.org/records/17845154

pweigmann:
> Question regarding the source: Is not this one the source we are relying on to get the full picture of emissions used in ScenarioMIP for CMIP7, which goes beyond CEDS?
> 
> https://zenodo.org/records/17845154

indeed, I must have mixed up the links. this is where the data is coming from. Regarding the name, I chose "CEDS_CMIP7" as most people referred to the data as CEDS data, even though the data for burning variables comes from another model as well.

edit: fixed the link. would you prefer a different name for the model?

R/calcEmissions.R
pweigmann: tested that it works for different datasources with https://github.com/pik-piam/mrremind/pull/813/commits/6070d820b5bb8afa53172cfd7517bb7d79bc6acd

fbenke-pik (approved):

pweigmann:
thanks for the helpful review @fbenke-pik !

laurinks:
> edit: fixed the link. would you prefer a different name for the model?

I was thinking about it. Options could be:

- `CMIP7_SMIP`
- `CMIP7_ScenarioMIP`
- `CMIP7_SMIP_CEDS`
- `CMIP7_SMIP_CEDSext`
- `CMIP7_CEDS`
- `CMIP7_CEDSext`

I have a preference for putting CMIP7 first. Otherwise, all options seem fine. Still, this data is specifically derived for the ScenarioMIP part of CMIP7, so having SMIP / ScenarioMIP in might be good.

I would opt for `CMIP7_SMIP_CEDSext` or `CMIP7_SMIP` depending on brevity preferences.

laurinks:
Thanks a lot, @pweigmann, for the very helpful work!

I went over the mapping once more and would actually suggest some changes to the naming of some sectors, in order to be more closely aligned with the AP from MAgPIE update that @dklein-pik is working on: https://github.com/remindmodel/development_issues/issues/590#issuecomment-3749108226

Here a concrete suggestion, also attached as csv file. What do you think? Do you see disadvantages?
[mappingCEDS_CMIP7.csv](https://github.com/user-attachments/files/28095994/mappingCEDS_CMIP7.csv)


image

pweigmann:
> Thanks a lot, @pweigmann, for the very helpful work!
> 
> I went over the mapping once more and would actually suggest some changes to the naming of some sectors, in order to be more closely aligned with the AP from MAgPIE update that @dklein-pik is working on: [remindmodel/development_issues#590 (comment)](https://github.com/remindmodel/development_issues/issues/590#issuecomment-3749108226)
> 
> Here a concrete suggestion, also attached as csv file. What do you think? Do you see disadvantages? 

Great, thanks!
- ``Industry`` does exist as sum of energy demand industry and industrial processes
- will change the transport variables
- on the burning variables: happy to change to these versions, I would just remove the "+" as we usually do it in the historical.mif, especially because I am not sure about the summations here. (side note: I also found ``Land|Biomass Burning|+|Agricultural Waste Burning/Savannah Fires...`` in older CEDS data in the historical.mif but this might be outdated)

laurinks:
> on the burning variables: happy to change to these versions, I would just remove the "+" as we usually do it in the historical.mif, especially because I am not sure about the summations here. (side note: I also found Land|Biomass Burning|+|Agricultural Waste Burning/Savannah Fires... in older CEDS data in the historical.mif but this might be outdated)

Yes, I am aware of that. There seems to be no perfect choice. My preference is to choose the names suggested above since they are close to the naming conventions in the ScenarioMIP context, and thus hopefully least confusing.

pweigmann:
> > on the burning variables: happy to change to these versions, I would just remove the "+" as we usually do it in the historical.mif, especially because I am not sure about the summations here. (side note: I also found Land|Biomass Burning|+|Agricultural Waste Burning/Savannah Fires... in older CEDS data in the historical.mif but this might be outdated)
> 
> Yes, I am aware of that. There seems to be no perfect choice. My preference is to choose the names suggested above since they are close to the naming conventions in the ScenarioMIP context, and thus hopefully least confusing.

Oh, and I think we will have some overlap in the transport variables if we change to the proposed variables. ~~Which will be relevant for summing to totals per species (counting domestic aviation twice).~~
Which means we will still not have a precise fit for the non-aviation/shipping transport variable.

laurinks:
After checking the XLSX, here are my notes:

1. `Energy and Industrial Processes`: I am not 100% sure about our current convention, but I thought that we typically have three variables: `Energy and Industrial Processes`, `w/ Bunkers|Energy and Industrial Processes`, and `w/o Bunkers|Energy and Industrial Processes`, where the first two are equal. The current version has 1 and 3, which is fine, but wanted to mention it.
2. There are the 4 `Emissions` variables at the end (`Emissions|CO2|AFOLU`, COE|EIP, GHG, and Kyoto GHG) and I am wondering how to best deal with these. Should we also call them `Emi`, possibly with some supplement ?
3. There is one big challenge with `CO2` (and thus also GHG): While there exist the variables Forest Burning, Grassland Burning, Peat Burning, and Agricultural Waste Burning, these are **not** taken into account when computing the total CO2 emissions that enter the simple climate models. As a consequence, we currently do not have `Emi|CO2` = `Emissions|CO2|AFOLU` + `Emissions|CO2|Energy and Industrial Processes`.  Should be briefly exchange how the aggregates are derived in mrremind to decide on a pragmatic way forward?

pweigmann:
Changes made:
- Renamed model to ``CMIP7_CEDS``.
- ``Emi|CMIP7|...`` instead of ``Emissions|...`` for reference variables. ``w/o AFOLU`` instead of ``Energy and Industrial Processes`` as it describes CMIP7 variable better.
- Calculated total now called ``Emi|<species>|w/o AFOLU`` and not containing burning variables any longer.
- Mapping adjusted to use better variable names for burning and transport vars
- Expanded and renamed new tool function, replacing custom aggregation functions in two more cases @fbenke-pik 

Am currently running fullVALIDATIONREMIND with these changes to see if it all works as expected.

The new file output @laurinks 
[historical_CEDS.xlsx](https://github.com/user-attachments/files/28142034/historical_CEDS.xlsx)

pweigmann:
Test historical.mif looks good: ``inputdata/output_1.27/rev7.88CMIP7_62eff8f7_validationremind.tgz``
One more change to aggregations: burning variables are added to totals for all species, except CO2.

laurinks:
Perfect, thank you very much @pweigmann !

PR #814, opened 2026-05-21
Weakening EU 2030 FE renewable share target

fschreyer:
This addresses https://github.com/remindmodel/development_issues/issues/734. 

There is a new method to determine the renewable FE share targets for 2030 in the EU as applying the full target from PBL was led to too extreme / non-credible EU transformation up to 2030

* FE RE share targets for EU in 2030 are set to national estimates / NCEPs as compiled [here](https://www.contexte.com/eu/article/energy/whos-ahead-whos-behind-eu-member-states-progress-on-renewables_225474)
* in addition there is another criterion: EU countries can only increase their RE share at 1.5 or 1 times their 2019-2024 speed (%/yr increase). The 1.5 or 1 depends on countries and whether they already grew fast or not or have a specific preference for nuclear (France).


R/calcRenShareTargets.R
```diff
@@ -307,23 +346,46 @@ calcRenShareTargets <- function(scenario) {
...
   # select which REMIND regions get targets and determine their respective target year
   x <- .selectTargetRemindRegion(x, xIntp, xHistTotal, regionmapping, 0.2)
 
-  ## 5. format renewable share target data to be used as input data for REMIND
+  ## 7. format renewable share target data to be used as input data for REMIND
 
   # rename target types to those used in REMIND techpol realization
-  getNames(x) <- c("RenElec", "RenFE", "NonBioRenElec", "NonFossilElec")
+  targetMapping <- data.frame(
+    source = c(
```
RahelMA, line 387: In Remind, I also added the type SE|solarwind as SolarWindElec. I wonder if we should also add the possible type here, as I remember that some policy targets cover solar and wind only.

RahelMA (approved):
Thank you, Felix, for adding the country granularity, thats really nice!

I only have a small comment on the types!

R/calcRenShareTargets.R
```diff
@@ -281,19 +283,56 @@ calcRenShareTargets <- function(scenario) {
   x <- collapseDim(x[, , "Min"])
 
 
-  ## 2. get historic renewable shares and totals ----
+  ## 2. make country-specific assumptions deviating from PBL input ----
```
lea-hayez, line 286: I know you are more keen on code comment rather than maintaining the Excel input sheet, but I still updated the EU ren share target in the NPi2025_PBL input file as "Adjusted" :)

fschreyer: I updated the calculation again now.  Modified assumption now is

EUR -> 31% in 2030 (instead of 42.5% before)

I can also give you per EU country if needed / of interest.

R/calcRenShareTargets.R
fschreyer: Let's add it once we read targets for this in via mrremind as well. I don't see the need to add a category here if we don't have data for it so far.

robertpietzcker:
super interesting, @fschreyer 

> EUR -> 31% in 2030 (instead of 42.5% before)
I can also give you per EU country if needed / of interest.

wow, that is a HUGE decrease! 

I would be grateful if you could you add a table with the per-country values as comment to this PR, so that it is all in one place?

robertpietzcker:
> I would be grateful if you could you add a table with the per-country values as comment to this PR, so that it is all in one place?

ok, I just saw that the change file directly contains the country-level percentage data, and it is not read in from some source file - so no need to copy this in here

PR #820, opened 2026-06-09
Nuclear and CCS Thresholds Update

pweigmann:
Update to the IEA CCUS data from 03.2026 and IAEA PRIS data from 08.06.2026. Also cleaned up ``calcProjectPipelines`` a bit and removed 2020 data from being written into the ``threshold.mif`` as we compare to historical.mif data anyway instead of using direct thresholds for past periods.

Overall only minor changes in values, but maybe relevant for REMIND input data @amerfort ? Find a before-after comparison in the attached files:
[thresholds_H12_2026-06-09_beforeCCSupdate.csv](https://github.com/user-attachments/files/28760562/thresholds_H12_2026-06-09_beforeCCSupdate.csv)
[thresholds_H12_2026-06-09_afterCCSupdate.csv](https://github.com/user-attachments/files/28760561/thresholds_H12_2026-06-09_afterCCSupdate.csv)


fbenke-pik (approved):

amerfort:
Thanks Pascal.
This is just for the validation tool, or do you also change the near-term bounds that will be used by REMIND as input data?
For V3.8.0 we then also have to update the REMIND near-term bounds to the new IEACCUS Database.

Looking at your comparison data:
Why are there NAs in some regions' red limits? And do you understand why the old data had such high yellow max limits for EUR? Comparing that with our REMIND upper limit for 2030, the old yellow max seems excessive. The new data with 62 Mt is more aligned with the REMIND bound logic. 

Once I'm back I'd like to understand whether we have inconsistent narratives between the validation tool and the REMIND bounds on CCS or if the project announcements in EUR declined so massively.

pweigmann:
Thanks for your input @amerfort ! I updated the data from IEA CCUS which is used by the validation tool and also the input data, but with different factors/methods. Merging this will affect REMIND input data CCS bounds.

I could have made clearer that the files I posted are only for use in the validation thresholds and don't show the REMIND bounds. It was the data I have at hand, a proxy to show how the underlying IEA CCUS data has changed overall within the last year, so best to look at relative changes between the two files.

Generally, I think we can continue with the approach we currently have where validation bounds and REMIND bounds don't follow the same logic (e.g. missing red bounds for regions for validation, other factors on announced projects) and generally the validation is less strict than REMIND.

Do you think these relative changes are too big to merge without understanding better what it will do to REMIND bounds and scenarios?

> For V3.8.0 we then also have to update the REMIND near-term bounds to the new IEACCUS Database.

This is what I do here, were you planning on also revisiting the method on how to use the data? Would you prefer to do this only for v3.8.0?

amerfort:
Alright, thanks for the clarification Pascal.
We should keep in mind that Europe's 2030 value decreased quite a bit and this will have an influence on EUR scenarios and Carbon prices.
But if the project prospect is less optimistic now, I think we should go ahead and include that in REMIND too.
Can you generate the new REMIND input data files for comparison too?
Thanks for the work and keeping me in the loop :)

amerfort (approved):
Thanks for the update

PR #826, opened 2026-06-24
add industry total emissions to EDGARghg output to historical.mif

robertpietzcker:
The EDGARghg output was missing "Industry", the sum of "Energy|Demand|Industry" and "Industrial Processes" - this is now added

( I am currently still doing the buildLibrary stuff)

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '56082094'
+ValidationKey: '559252190'
 AcceptedWarnings:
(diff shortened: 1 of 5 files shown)


robertpietzcker:
when doing buildLibrary, suddenly many .Rd files change: 

image

because `magclass:magclass` is replaced by `magclass:magclass-package`

Is this fine and should I simply commit all those 26 changed files?

fbenke-pik (approved):

PR #827, opened 2026-06-25
New biomass supply curves

dklein-pik:
Apply new biomass supply curves derived from MAgPIE 4.14.0.
Needs `mrcommons` >= 1.70.0


LaviniaBaumstark (approved):
code changes look good, why are so many *.Rd files changed?

PR #805, opened 2026-04-13
incorporate new buildings ces tree and expand buildings input variables

hagento:
In this PR, I adapted the buildings input data to REMIND's new CES tree structure and to provide disaggregated carrier-enduse pair data from EDGE Buildings.

**Key changes:**
- Updated buildings variable mappings in`mappingEDGEBuildingsToREMIND.csv` to align with REMIND's new CES tree structure.
- Extended output to include full disaggregated final energy data for all carrier-enduse pairs from EDGE Buildings alongside aggregated variables. This disaggregated data will be needed for future post-processing procedures and planned feature enhancements on endogenous climate impacts.

**New Buildings CES Structure**

```mermaid
graph TD
    en["en (total energy)"]
    enb["enb (buildings energy)"]
    enhb["enhb (buildings heating energy)"]
    enhgab["enhgab (gases)"]

    fesob["fesob (solids)"]
    fehob["fehob (liquids)"]
    feheb["feheb (district heat)"]
    feelhpb["feelhpb (electricity for heat pumps)"]
    feelrhcob["feelrhcob (electricity for resistive heating and cooking)\n[RENAMED from feelrhb]"]

    fegab["fegab (natural gas)"]
    feh2b["feh2b (hydrogen)"]

    feelalb["feelalb (electricity for appliances and lighting)\n[NEW - was part of feelcb]"]
    feelictb["feelictb (electricity for ICT)\n[NEW - was part of feelcb]"]
    feelscb["feelscb (electricity for space cooling)\n[NEW - was part of feelcb]"]

    en --> enb
    enb --> enhb
    enb --> feelalb
    enb --> feelictb
    enb --> feelscb

    enhb --> fesob
    enhb --> fehob
    enhb --> feheb
    enhb --> feelhpb
    enhb --> feelrhcob
    enhb --> enhgab

    enhgab --> fegab
    enhgab --> feh2b
```

**Note**
Once EDGE-Buildings PR [#40](https://github.com/pik-piam/edgebuildings/pull/40) is merged, I will update the input revision in `R/readEdgeBuildings.R` and open this PR to merge.

PR #832, opened 2026-07-03
re-introduce non-specified buildings electricity demand

hagento:
In this PR I am fixing a bug introduced in PR #805 where I incorrectly removed the aggregation of non-specified / other electricity demand from the function `calcFeDemandONONSPEC` to non-heating buildings electricity demand (formerly `feelcb`). Given that the demand is not defined in its energy service, I decided to aggregate it to `feelalb` (fe electricity for appliances & lighting) since this is the most broad category we have. 

The bug was found when comparing input data from revisions 7.97 (before merge of PR #805) and 8.12 (most recent), when historical FE electricity demands for China deviated significantly. After the correction (8.12CHAelec), they now align nicely again. The deviation in future electricity projections comes from the introduction of exogenous ICT demand projections. 

CHA_elec

You can find the input data with revision `8.12CHAelec`.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '56230375'
+ValidationKey: '56256462'
 AcceptedWarnings:
 - Invalid URL: .*
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


robinhasse (approved):
Thanks Hagen!

PR #838, opened 2026-07-16
use new subtype of readStegmann2022 in calcPlasticsEoL

leonieschweiger:
see https://github.com/pik-piam/mrindustry/pull/75


fbenke-pik:
Don't forget to run buildLibrary

fbenke-pik (approved):

PR #840, opened 2026-07-24
Simplify calcUNFCCC (extract functions)

simonkrogmann:
Extract functions from calcUNFCCC to avoid duplicate code. Specifically the duplicate aggregation by sector seemed very error-prone to me when editing magic numbers.

### DIFF
--- a/R/calcUNFCCC.R
+++ b/R/calcUNFCCC.R
@@ -16,23 +16,24 @@ calcUNFCCC <- function(subtype = "all") {
 
   mapping <- toolGetMapping("Mapping_UNFCCC.csv", type = "reportingVariables", where = "mrremind") %>%
     select("variable" = "UNFCCC", "REMIND", "conversion" = "Factor", "Unit_REMIND") %>%
-    mutate("REMIND" = trimws(.data$REMIND),
-           "variable" = trimws(gsub("\\.", "_", .data$variable))) %>%
-    filter(!is.na(.data$REMIND), .data$REMIND != "")
+    mutate("REMIND" = trimws(.data[["REMIND"]]),
+           "variable" = trimws(gsub("\\.", "_", .data[["variable"]]))) %>%
+    filter(!is.na(.data[["REMIND"]]), .data[["REMIND"]] != "")
 
   df <- data %>%
-    mselect(variable = unique(mapping$variable)) %>%
-    quitte::as.quitte(na.rm = TRUE) %>%
-    filter(.data$period >= 1990)
+    mselect(variable = unique(mapping[["variable"]])) %>%
+    as.data.frame(rev = 3) %>%
+    filter(!is.na(.data[[".value"]])) %>%
+    filter(.data[["year"]] >= 1990)
 
   x <- left_join(df, mapping, by = "variable", relationship = "many-to-many") %>%
     mutate(
-      "value" = .data$value * .data$conversion,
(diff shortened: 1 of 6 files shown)


simonkrogmann:
@robertpietzcker What do you think of the interface for future data edits regarding the three functions I inserted?

robertpietzcker:
@simonkrogmann apologies, I only will have time to have a look at this in 4 weeks or so when I am back from vacations

fbenke-pik:
@pweigmann Simon rewrote some calculations to be less redundant. Would you say the new functions are easy enough to understand to enable a non-coder to make adjustments here?

fbenke-pik:
@simonkrogmann Please generate a cs4r file with the results of calcUNFCCC before and after your changes and verify that they are identical using `piamutils::compareMagpieObject`

R/calcUNFCCC.R
```diff
@@ -267,57 +221,37 @@ calcUNFCCC <- function() {
     x[, , "Emi|GHG|Industrial Processes (Mt CO2eq/yr)"] +
     x[, , "Emi|GHG|Energy|Demand|Industry (Mt CO2eq/yr)"]
 
+  .aliasColumn <- function(x, aliasName, existingColumn) {
+    stopifnot(existingColumn %in% getItems(x)[["variable"]])
```
pweigmann: > ``getItems(x)[["variable"]]``

this should probably better be ``getItems(x, dim = 3.1)``

pweigmann:
Thanks @simonkrogmann for cleaning this up! A few considerations:
- With magical numbers I assume you mean the GWPs? It is something that has bothering me and others for a while and there is actually a task to solve this properly with a little package which can provide GWPs for various species and sources (AR5, vs AR6 for example). The change here is certainly an improvement but we still have the bigger problem to make sure we actually use consistent GWPs across functions and packages. I started it [here](https://github.com/pweigmann/gwpdata) a long time ago but never got around to finishing it.
- The ``glue`` functions seems very useful, especially for people used to Python. I think it's good to use it here but generally we also should try to reduce the number of dependencies. What we would usually do and what people are used to, is
``paste0("Emi|GHG|", sector, " (Mt CO2eq/yr)")``
instead of
``glue::glue("Emi|GHG|{sector} (Mt CO2eq/yr)")``
I think it should be equivalent, right?
- extracting functions also makes sense, especially if we start doing it across other functions as well, so people get used to it. otherwise the line reduction might not be worth the added complexity and effort

R/calcUNFCCC.R
simonkrogmann: Done

simonkrogmann:
- I hope your unification of GWPs becomes easier with my change (less spots, easier to find).
- `glue` is already heavily used in mrremind, that's why I think it's acceptable and more readable but maybe I'm biased coming from Python. (Your code is equivalent.)
- My aim is to extract repetitive functions in many places but keep them relatively simple. (I hope I succeeded here.) I think they are badly needed, because they prevent some editing mistakes:
```
  x <- add_columns(x, "Emi|GHG|Energy|Demand (Mt CO2eq/yr)",
                   dim = 3)
  x[, , "Emi|GHG|Energy|Demand (Mt CO2eq/yr)"] <-
    x[, , "Emi|GHG|w/ Bunkers|Energy|Demand (Mt CO2eq/yr)"]
```
Here, the first two strings need to be identical and I forgot to update one in an intermediate version of https://github.com/pik-piam/mrremind/pull/841.

pweigmann (approved):

PR #841, opened 2026-07-27
Calc output for UNFCCC non-annex-data

simonkrogmann:
This maps the data read from non-annex-1 countries in https://github.com/pik-piam/mrcommonsenergy/pull/13 to remind variables. I also added aggregate variables for the annex data to match the non-annex data.

Additionally, I modified the handling of subdimensions.

### DIFF
--- a/R/calcEmiTargetReference.R
+++ b/R/calcEmiTargetReference.R
@@ -14,7 +14,7 @@ calcEmiTargetReference <- function() {
   gwpN2O <- 265
 
   ceds <- calcOutput("Emissions", datasource = "CEDS2025", years = 1990:2022, aggregate = FALSE)
-  unfccc <- collapseDim(calcOutput("UNFCCC", years = 1990:2022, aggregate = FALSE, warnNA = FALSE))
+  unfccc <- calcOutput("UNFCCC", years = 1990:2022, aggregate = FALSE, warnNA = FALSE)
 
   # calculate CEDS values ----
 
--- a/R/calcEmiTargetReference.R
+++ b/R/calcEmiTargetReference.R
@@ -14,7 +14,7 @@ calcEmiTargetReference <- function() {
   gwpN2O <- 265
 
   ceds <- calcOutput("Emissions", datasource = "CEDS2025", years = 1990:2022, aggregate = FALSE)
-  unfccc <- calcOutput("UNFCCC", years = 1990:2022, aggregate = FALSE, warnNA = FALSE)
+  unfccc <- calcOutput("UNFCCC", subtype = "annex-1-only", years = 1990:2022, aggregate = FALSE, warnNA = FALSE)
 
   # calculate CEDS values ----
 
--- a/R/calcUNFCCC.R
+++ b/R/calcUNFCCC.R
@@ -7,7 +7,7 @@
 #'
 calcUNFCCC <- function() {
 
-  data <- readSource("UNFCCC")
+  data <- readSource("UNFCCC", subtype = "annex-1")
 
   # fill countries of selected regions with 0 to allow for regional aggregation
   regions.fill <- c("EUR", "REF", "NEU", "CAZ")
@@ -41,12 +41,11 @@ calcUNFCCC <- function() {
     select("variable" = "REMIND", "region", "period", "value")
 
   x <- stats::aggregate(value ~ variable + region + period, x, sum) %>%
-    as.magpie() %>%
-    toolCountryFill(fill = NA, verbosity = 2)
+    as.magpie()
 
   # aggregate pollutants ----
 
-  x <- add_columns(x, "Emi|CH4|w/o Bunkers|LULUCF national accounting (Mt CH4/yr)", dim = 3.1)
+  x <- add_columns(x, "Emi|CH4|w/o Bunkers|LULUCF national accounting (Mt CH4/yr)", dim = 3)
   x[, , "Emi|CH4|w/o Bunkers|LULUCF national accounting (Mt CH4/yr)"] <- dimSums(
     x[, , c(
       "Emi|CH4|Agriculture (Mt CH4/yr)",
@@ -58,7 +57,7 @@ calcUNFCCC <- function() {
     dim = 3, na.rm = TRUE
   )
 
-  x <- add_columns(x, "Emi|CO2|w/o Bunkers|LULUCF national accounting (Mt CO2/yr)", dim = 3.1)
+  x <- add_columns(x, "Emi|CO2|w/o Bunkers|LULUCF national accounting (Mt CO2/yr)", dim = 3)
   x[, , "Emi|CO2|w/o Bunkers|LULUCF national accounting (Mt CO2/yr)"] <- dimSums(
     x[, , c(
       "Emi|CO2|Agriculture (Mt CO2/yr)",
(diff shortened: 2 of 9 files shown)


R/calcUNFCCC.R
```diff
@@ -324,7 +324,39 @@ calcUNFCCC <- function() {
 
   # remove years before 1990 due to incomplete data
   x <- x[, seq(1986, 1989, 1), , invert = TRUE]
-  x <- add_dimension(x, dim = 3.1, add = "model", nm = "UNFCCC")
+
+
+
+  # ==================== Non-Annex-1 Data ======================
```
fbenke-pik: Use `Non-Annex-1 Data ----` to make collapsable sections in RStudio (VScode as well?)

R/calcUNFCCC.R
```diff
@@ -324,7 +324,39 @@ calcUNFCCC <- function() {
...
+  nonAnnexData <- readSource("UNFCCC", subtype = "non-annex-1")
+
+  mapping <- toolGetMapping("Mapping_UNFCCC_Aggregate.csv", type = "reportingVariables", where = "mrremind") %>%
```
fbenke-pik: What does Aggregate stand for? I would choose a name indicating that this is the mapping for non-annex countries (and maybe also rename the annex mapping accordingly)

R/calcUNFCCC.R
```diff
@@ -324,7 +324,39 @@ calcUNFCCC <- function() {
...
+    quitte::as.quitte(na.rm = TRUE)
+
+  nonAnnexData <- left_join(nonAnnexData, mapping, by = "variable", relationship = "many-to-many") %>%
+    mutate(
+      "value" = .data$value * .data$conversion,
+      "REMIND" = paste0(.data$REMIND, " (", .data$Unit_REMIND, ")")
+    ) %>%
+    select("variable" = "REMIND", "region", "period", "value")
+
+  nonAnnexData <- stats::aggregate(value ~ variable + region + period, nonAnnexData, sum) %>%
```
fbenke-pik: Technically, aggregate is not necessary in your case, as you only have 1:1 mappings but if you prefer, feel free to leave it as it is.

fbenke-pik (approved):

fbenke-pik:
For additional validation, make sure that your changes don't introduce new warnings and the resulting magpie object has the correct regions, years and additional variables.

R/calcUNFCCC.R
simonkrogmann: Replaced it with Non_Annex

R/calcUNFCCC.R
simonkrogmann: Done

R/calcUNFCCC.R
simonkrogmann: I removed it

inst/extdata/reportingVariables/Mapping_UNFCCC_Non_Annex.csv
```diff
@@ -0,0 +1,7 @@
+UNFCCC;Factor;REMIND;Unit_REMIND
+CO2 emissions without LULUCF / LUCF;0.001;Emi|CO2|w/o Bunkers|w/o Land-Use Change;Mt CO2/yr
+CO2 net emissions/removals by LULUCF / LUCF;0.001;Emi|CO2|Land-Use Change|LULUCF national accounting;Mt CO2/yr
+CO2 net emissions/removals with LULUCF / LUCF;0.001;Emi|CO2|w/o Bunkers;Mt CO2/yr
```
fschreyer: Sry, I realize that this mapping is not correct. As UNFCCC always uses LULUCF national accounting and that is not the REMIND default we need to map it to different variable names including ``|LULUCF national accounting`` in the end. 

So, it needs to be ``Emi|CO2|w/o Bunkers|LULUCF national accounting``. The same for ``Emi|GHG|w/o Bunkers``. It should rather be ``Emi|GHG|w/o Bunkers|LULUCF national accounting``.

R/calcUNFCCC.R
```diff
@@ -84,250 +83,325 @@ calcUNFCCC <- function() {
...
     x[, , "Emi|CO2|Energy|Demand|Industry|Other Industry (Mt CO2/yr)"]
 
-  x <- add_columns(x, "Emi|GHG|Industry (Mt CO2eq/yr)", dim = 3.1)
+  x <- add_columns(x, "Emi|GHG|Industry (Mt CO2eq/yr)", dim = 3)
   x[, , "Emi|GHG|Industry (Mt CO2eq/yr)"] <-
     x[, , "Emi|GHG|Industrial Processes (Mt CO2eq/yr)"] +
     x[, , "Emi|GHG|Energy|Demand|Industry (Mt CO2eq/yr)"]
 
+  # overall aggregates ----
+  x <- add_columns(x, "Emi|CO2|w/o Bunkers (Mt CO2/yr)", dim = 3)
```
fschreyer: As stated above. All aggregate variables that contain ``Emi|XXX|Land-Use Change|LULUCF national accounting`` should be get a ``|LULUCF national accounting`` at the end.

R/calcEmiTargetReference.R
```diff
@@ -14,7 +14,7 @@ calcEmiTargetReference <- function() {
   gwpN2O <- 265
 
   ceds <- calcOutput("Emissions", datasource = "CEDS2025", years = 1990:2022, aggregate = FALSE)
-  unfccc <- collapseDim(calcOutput("UNFCCC", years = 1990:2022, aggregate = FALSE, warnNA = FALSE))
+  unfccc <- calcOutput("UNFCCC", years = 1990:2022, aggregate = FALSE, warnNA = FALSE)
```
fschreyer: Is the output of calcEmiTargetReference changing with your PR? Does it affect the output of ``convertUNFCCC_NDC``?

This is exactly the function where I would like to use these data. But I would not like to change it right away. Would be good to make sure that the non-Annex I UNFCCC data does not affect the calcEmiTargetReference output so far until we review the whole calculation of NDC targets in ``convertUNFCCC_NDC``, ``calcEmiTargetReference`` etc.

fschreyer (changes requested):
Thanks! I realized that some variable names should be mapped differently. See comments below.

R/calcUNFCCC.R
simonkrogmann: `Emi|GHG|w/o Bunkers|LULUCF national accounting` and `Emi|CO2HG|w/o Bunkers|LULUCF national accounting` are duplicate then, so I removed the newly inserted one

R/calcEmiTargetReference.R
simonkrogmann: I added a subtype "annex-1-only" to mitigate this.

R/calcUNFCCC.R
```diff
@@ -84,250 +78,316 @@ calcUNFCCC <- function() {
...
+      select("variable" = "UNFCCC", "REMIND", "conversion" = "Factor", "Unit_REMIND") %>%
+      mutate(
+        "REMIND" = trimws(.data[["REMIND"]])
+      ) %>%
+      filter(!is.na(.data[["REMIND"]]), .data[["REMIND"]] != "")
+
+    nonAnnexData <- nonAnnexData %>%
+      mselect(variable = unique(mapping[["variable"]])) %>%
+      as.data.frame(rev = 3) %>%
+      filter(!is.na(.data[[".value"]]))
```
fbenke-pik, line 346: are you sure that he column is called `.value` and not `value`?

simonkrogmann: Yes, magpie to dataframe conversion produces peculiar column names. Even stranger when I used the default, rev = 1

R/calcUNFCCC.R
```diff
@@ -84,250 +78,316 @@ calcUNFCCC <- function() {
...
+
+    years = union(getYears(x), getYears(nonAnnexData))
+    names = union(getNames(x), getNames(nonAnnexData))
+
+    # write both annex and non-annex data into a new magclass object
+    result <- new.magpie(
+      cells_and_regions = getISOlist(),
+      years = years,
+      names = names,
+      sets = c("region", "year", "value"),
```
fbenke-pik, line 368: Calling the 3rd dimension will cause problems later on. Inputdata generation in fullVALIDATION crashes when appending the magpie object to historical.mif  Just name the 3rd dimension `variable` as `value` is kinda reserved for the actual values.

`"~ Argument 'value.name' should not overlap with column names in result: [Region, year, value]"`

PR #842, opened 2026-07-31
Fix column name conflict

simonkrogmann:
Calling the dimension of a magpie return object `value` clashes with internals of madrat (specifically the `file` option of `calcOutput`). This PR renames it to avoid the conflict.


fbenke-pik (approved):
Thank you!

PR #843, opened 2026-08-05
update BP (Energy Institute) reference data

pweigmann:
Update to the 2026 release of the Statistical Review of World Energy by Energy Institute (formerly BP) for use in the historical.mif.

Mostly updating to include new years and adjusting workbook ranges. One methodological change is that Total Energy Supply (TES) is now directly reported as electrical output, only nuclear still needs conversion in ``calcBP``. Minor changes to mappings/names.


fbenke-pik (approved):
Thanks for the update! I am pleasantly surprised that they kept the formats fairly consistent (judging from the amount of necessary changes).

